PR #7, opened 2017-06-29
Cache responsibility sets

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-360


app/src/main/contrib/models/ResponsibilitySet.ts
```diff
@@ -48,4 +51,9 @@ export class ExtendedResponsibilitySet implements IExtendedResponsibilitySet {
         const responsibility = this.getResponsibilityByScenario(scenarioId);
         responsibility.coverageSets = coverageSets;
     }
+
+    getCoverageSets(scenarioId: string): CoverageSet[] {
+        const responsibility = this.getResponsibilityByScenario(scenarioId);
+        return responsibility.coverageSets;
+    }
```
littledice: I'm a bit lost - do these changes relate to the caching story?

MartinEden:
This should only be reviewed and merged once #8 and #9 are done.

What's going on here is that the responsibility store only stored one responsibility set at a time - the one you were currently looking at. In order to cache things I had to restructure the store so that it could store all the responsibility sets for all touchstones and all modelling groups.

app/src/main/contrib/stores/ResponsibilityStore.ts
```diff
@@ -96,30 +112,37 @@ class ResponsibilityStore extends AbstractStore<ResponsibilityState, Responsibil
     }
 
     handleSetCurrentResponsibility(scenarioId: string) {
-        if (this.responsibilitySet != null) {
-            this.currentResponsibility = this.responsibilitySet.getResponsibilityByScenario(scenarioId);
+        const instance = this.getInstance();
+        const set = this.getInstance().getCurrentResponsibilitySet();
+        if (set != null) {
```
littledice: remove instance const

app/src/test/contrib/stores/ResponsibilityStoreTests.ts
```diff
@@ -75,19 +74,24 @@ describe("ResponsibilityStore", () => {
         expect(responsibilityStore.getState().currentModellingGroup).to.eql(group);
     });
 
-    it("responsibilityActions.update sets responsibility set", () => {
+    it("responsibilityActions.update sets current responsibility set", () => {
```
littledice: extra word

MartinEden: Actually we were both confused here, but I think the code is right as it stands. We just confused the action name with the sentence.

MartinEden:
All done here.

PR #10, opened 2017-06-29
Refactor

littledice:
Refactored to share `ModellingGroupSource` and `ModellingGroupActions` between the apps.


littledice:
It has become more difficult to share this code now that caching is introduced, as the previously state independent call to `fetchModellingGroups` now needs to check state in `isCached`

PR #11, opened 2017-06-29
I159

littledice:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-159

### DIFF
--- a/.idea/vcs.xml
+++ b/.idea/vcs.xml
@@ -1,6 +1,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<project version="4">
-  <component name="VcsDirectoryMappings">
-    <mapping directory="" vcs="Git" />
-  </component>
-</project>
--- a/.idea/vcs.xml
+++ b/.idea/vcs.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="VcsDirectoryMappings">
+    <mapping directory="" vcs="Git" />
+  </component>
+</project>
--- a/app/src/main/admin/actions/UserActions.ts
+++ b/app/src/main/admin/actions/UserActions.ts
@@ -2,6 +2,28 @@ import { alt } from "../../shared/alt";
 import { FetchActions, FetchActionsInterface } from "../../shared/actions/FetchActions";
 import { User } from "../../shared/models/Generated";
 
-export const userActions = alt.createActions<FetchActionsInterface<User[]>>(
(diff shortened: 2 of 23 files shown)


.idea/vcs.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="VcsDirectoryMappings">
+    <mapping directory="" vcs="Git" />
+  </component>
+</project>
```
MartinEden: I'd rather we excluded .idea files for the repository, even if they are unlikely to differ between users.

app/src/main/admin/actions/UserActions.ts
```diff
@@ -2,6 +2,28 @@ import { alt } from "../../shared/alt";
...
 
-export const userActions = alt.createActions<FetchActionsInterface<User[]>>(
-    class extends FetchActions<User[]> {}
-);
+interface Actions extends FetchActionsInterface<User[]> {
+    setCurrentUser(username: string): string;
+    beginFetchDetails(username: string): string;
+    updateUserDetails(details: User): User;
+
+}
```
MartinEden: I've been trying to move away from using `FetchActionsInterface` if there is more than one set of `beginFetch` and `update` on the class.

So rather than having the slightly confusing: `beginFetch`, `update`, `beginFetchDetails`, `updateDetails`, I'd prefer to have `beginFetchUsers`, `updateUsers`, `beginFetchDetails`, `updateDetails`.

app/src/main/admin/components/Users/SingleUser/UserDetailsContent.tsx
```diff
@@ -0,0 +1,62 @@
...
+        if (props.user.roles != null)
+            roles =
+                <tr>
+                    <td>Roles</td>
+                    <td>
+                        {props.user.roles
+                            .map(r =>
+                                <UserRole key={ r.name + r.scope_prefix + r.scope_id } {...r} />)}
+                    </td>
+                </tr>;
```
MartinEden: I'm generally not hopping keen on single line `ifs` with omitted braces, but I haven't been jumping on them in your code, because we don't have to agree on every code style point. But I really don't think this is clear - much better to have braces around this `if` statement.

Also, maybe separate out the role rendering logic into another method?

littledice:
Yeah I was following the groups logic and kinda forgot that you get the same level of detail back from the `/users/` endpoint as from the `/user/:username` one

littledice:
All PR comments addressed

PR #14, opened 2017-07-20
I417

littledice:
page with list of reports and for a chosen report, page with list of versions


app/src/main/report/sources/ReportingSource.ts
```diff
@@ -0,0 +1,34 @@
...
+import {ReportingFetcher} from "./ReportingFetcher";
+import fetcher from "../../shared/sources/Fetcher";
+
+export abstract class ReportingSource<TState> extends Source<TState> {
+
+    protected doFetch<TModel>(urlFragment: UrlBuilder<TState>, config: FetchConfig<TState, TModel>): AltJS.SourceModel<TModel> {
+        const handler = this.processResponse;
+        const source: AltJS.SourceModel<TModel> = {
+            remote(state: TState) {
+                return handler((<ReportingFetcher>fetcher.fetcher).fetchFromReportingApi(urlFragment(state)))
```
littledice: the only difference between this and the Source base class is this line, that casts the fetcher to a ReportingFetcher, and calls through to the Orderly API instead of the Montagu API.

app/src/test/report/fetch/ReportingFetchHelper.ts
```diff
@@ -0,0 +1,162 @@
...
+
+
+function mockReportingFetcher(promise: Promise<Response>) {
+    fetcher.fetcher = new ReportingFetcher();
+    fetcher.fetcher.fetchFromReportingApi = function(urlFragment: string, options?: FetchOptions, includeToken: boolean = true) {
+        return promise;
+    };
+}
+
+export class ReportingFetchHelper<TPayload> {
```
littledice: this class is just like the FetchHelper class, except it instantiates the fetcher as a ReportingFetcher and spies on "fetchFromReportingAPI" instead of "fetch"

app/src/main/report/README.md
```diff
@@ -0,0 +1,12 @@
+# Montagu Reporting portal
+This portal allows funders (and anyone else) to access the reports in the Montagu Reports Orderly store.
+
+# Development
+1. Run `npm install` to get dependencies and `npm install webpack --global` to install webpack
+2. Run `webpack` to build, or `webpack --watch` to continuously monitor files and
+rebuild as needed.
+3. You will need to run both the Montagu API and the Montagu Reporting API using a shared RSA keypair. 
+Each of these contain a script called `runAPIForWebappDev` and running both of these in either order will set everything up. 
+Note that these scripts need to be run as `sudo` and will generate a keypair in `/etc/montagu/api/token_key.`
```
MartinEden: Once you are using the cert tool to generate the keypair (over in the API repo) then you could actually directly map `/etc/montagu/api/token_key/` as the dir (volume) to output to. Then you wouldn't need sudo to run this (just membership of the docker group).

app/src/main/report/sources/ReportingSource.ts
MartinEden: Would it be possible to just extend from Source base class and override one part of it? I know there's been a lot of copy paste in the API side of things, but that's a bit different as its a whole new code base.

app/src/test/report/fetch/ReportingFetchHelper.ts
MartinEden: Again, a lot of duplication that it would be better to avoid. Maybe you could avoid the duplication if the ReportFetcher had the same interface as Fetcher. i.e. `fetch`, rather than `fetchFromReportingApi`?

littledice: yeah the reason I didn't do that is because it also uses the original 'fetch' to get the auth token

app/src/main/report/sources/ReportingSource.ts
littledice: thats exactly what it is doing - it extends Source and overrides this one method

app/src/test/report/fetch/ReportingFetchHelper.ts
littledice: the alternative would be to re-write/duplicate the LoginForm instead... the amount of duplication would be roughly similar

app/src/main/report/README.md
littledice: Sounds good, have added note to relevant card to update this README when done

PR #27, opened 2017-08-17
Portal allows user to select gender option for dataset where available

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-462

I can't prove this works, because of the bug with VIMC-461, but it seems to generate the right URLs. Up to you whether you review and merge it now or later.


littledice:
The only thing that needs to change to work with the real endpoint is that the gender codes should be single capital letters.

I.e. `{value: "both", label: "Both" }` should become {` value: "B", label: "Both" }`

MartinEden:
I thought we discussed that it would be better if the API expected "female", "male" and "both" rather than the more cryptic single letter codes? Could we go down that path?

littledice:
Oh, yes we could. I was inferring from [this ticket](https://vimc.myjetbrains.com/youtrack/issue/VIMC-496)  that we were using codes rather than full names, but no real reason we should.

PR #34, opened 2017-08-24
Make URL match that generated by API email

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-582


MartinEden:
I also tidied up URLs so that they all terminate in a /, for consistency.

PR #40, opened 2017-09-05
Reporting portal is returning only first artefact of each type

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-608

### DIFF
--- a/app/src/main/report/components/Artefacts/ArtefactsList.tsx
+++ b/app/src/main/report/components/Artefacts/ArtefactsList.tsx
@@ -4,6 +4,8 @@ import {settings} from "../../../shared/Settings";
 import {Artefact} from "../../../shared/models/reports/Artefact";
 import {ArtefactItem} from "./Artefact";
 
+const styles = require("../../styles/reports.css");
+
 interface ArtefactsListProps{
     artefacts: ILookup<Artefact>[],
     report: string,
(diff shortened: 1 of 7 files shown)


app/src/test/report/components/Artefacts/ArtefactsListTests.tsx
```diff
@@ -6,25 +6,31 @@ import {mockArtefact} from "../../../mocks/mockModels";
 
 describe("ArtefactsList", () => {
     it("renders list of ArtefactItems", () => {
-
         const fakeArtefactsArray = [
             {
-                "staticgraph": mockArtefact({filename: "file.csv", description: "a file"})
+                "staticgraph": mockArtefact({
```
richfitz: If there's any way of harmonising the test data orderly produces and the information here, that might help keep things in sync more easily

app/src/main/report/components/Artefacts/ArtefactsList.tsx
```diff
@@ -18,13 +20,15 @@ export class ArtefactsList extends React.Component<ArtefactsListProps, undefined
     }
 
     render() {
-
         const artefactItems = this.props.artefacts
-            .map((artefact)=> {
-                const key = Object.getOwnPropertyNames(artefact)[0];
-                return <ArtefactItem key={key} report={this.props.report} version={this.props.version} {...artefact[key]} />;
+            .map((artefactLookup)=> {
```
richfitz: I presume this is the bulk of the actual code change - at some point it would be good if someone could walk me a little through how this is all put together as this mostly shows me I have no idea how things are being done

richfitz: (that's not at all a criticism of the code! Just me not knowing how this is all put together)

MartinEden: To be honest, this code is fairly hard to read, owing the unusual JSON format Orderly is using. Data in the format you suggested as alternative in the ticket would be a lot more normal:

```
[
    {
        "type": "data",
        "filenames": [
            "all.csv"
        ],
        "description": "raw export"
    },
    {
        "type": "data",
        "filenames": [
            "subset.csv"
        ],
        "description": "the subset we care most about"
    },
    ...
]
```

Then this could read something like:

```
const artefactItems = this.props.artefacts
            .map(artefact => {
                return <ArtefactItem key={artefact.description} report={this.props.report} version={this.props.version} {...artefact} />;
            });
```

richfitz: It's more that I don't have a model of what is going on in general (in a language I've never used).  But yes, that does look easier to read too.

One option that @littledice suggested was that we could move from storing json to working with relational data in the SQLite database.  That might be simpler than dealing with json strings.  I can discuss that with her when she's back perhaps

PR #43, opened 2017-09-07
Turn off caching.

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-578

This doesn't actually disable caching; it just ensures that clients check with the server what the latest version is before deciding whether to use their local cached copy or the newer one.


MartinEden:
That's just how web clients / web servers work by default. For ngninx serving static files this will be happening automatically. And for the dynamically generated API content, it will never cache. (We might add more intelligent caching later if we hit performance problems).

PR #44, opened 2017-09-12
I619

littledice:
* Component to display links to burden estimate templates.
* Templates themselves to be named using the disease id and put into folder `src/templates` that will be copied across to the docker container to be served statically by nginx.

nb. waiting on remaining templates


littledice:
How about this - group id prefixed template names?

PR #56, opened 2017-10-06
Stacking error messages

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-615

### DIFF
--- a/app/src/test/shared/actions/NotificationActionTests.ts
+++ b/app/src/test/shared/actions/NotificationActionTests.ts
@@ -6,12 +6,12 @@ import {
 } from "../../../main/shared/actions/NotificationActions";
 
 describe("NotificationActions", () => {
-    it("does not change notification", () => {
+    it("notify passes through notification unchanged", () => {
         const notification = makeNotification("message", "info");
         expect(notificationActions.notify(notification)).to.eql(notification);
     });
 
-    it("converts NotificationException into notification", () => {
+    it("notify converts NotificationException into notification", () => {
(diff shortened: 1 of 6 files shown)


app/src/test/shared/actions/NotificationActionTests.ts
```diff
@@ -6,12 +6,12 @@ import {
...
+    it("notify passes through notification unchanged", () => {
         const notification = makeNotification("message", "info");
         expect(notificationActions.notify(notification)).to.eql(notification);
     });
 
-    it("converts NotificationException into notification", () => {
+    it("notify converts NotificationException into notification", () => {
         const exception = makeNotificationException("message", "info");
         expect(notificationActions.notify(exception)).to.eql({
             message: "message",
```
MartinEden: Renamed these two tests because it took me a while to figure out what they did and whether they needed updating!

app/src/test/shared/components/ErrorLogTests.tsx
```diff
@@ -3,19 +3,32 @@ import { shallow } from "enzyme";
 import { expect } from "chai";
```
MartinEden: All that's happened here is there is one new test at the end, but the whitespace was messed up (indented with 3 spaces instead of 4) so everything has changed.

MartinEden:
I'd like to make that a separate ticket, as it's quite complicated. I've raised https://vimc.myjetbrains.com/youtrack/issue/VIMC-787 for this. That way we can merge this in and keep making progress.

PR #57, opened 2017-10-10
get latest bootstrap with npm

littledice:
and serve in document head

### DIFF
--- a/app/package-lock.json
+++ b/app/package-lock.json
@@ -1156,6 +1156,11 @@
         "hoek": "2.16.3"
       }
     },
+    "bootstrap": {
+      "version": "4.0.0-beta",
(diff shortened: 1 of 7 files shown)


littledice:
V valid question :) I think best to get ahead and use the latest version to avoid becoming quickly out of date; the Bootstrap blog suggests there will be no breaking changes between beta and production v4.

http://blog.getbootstrap.com/2017/08/10/bootstrap-4-beta/
"Long story short, shipping a beta means we’re done breaking all your stuff until our next major version (v5). We’re not perfect, but we’ll be doing our best to keep all the classes, features, and docs URLs as they appear now in this release. We can always add more things, but we cannot take away."

PR #58, opened 2017-10-10
new templates for Xi Li

littledice:


littledice:
Yeah this needs updating

MartinEden:
I'm afraid this still doesn't seem right. In the new version of the CSV I see "hepb_cases_acute_fulminant", but not in the science DB. It's possible that the science DB is out of date, and there is an another database PR (other than https://github.com/vimc/montagu-db/pull/23) that adds these missing outcomes? If so, could you please link to it? Sorry if I've already seen that and just forgotten it!

littledice:
No you're right, I don't see that in the db either. Sorry for missing that, will follow up.

PR #59, opened 2017-10-17
Remove all references to deprecated fields

littledice:
* Updated models to point to master, and re-generated classes
* Removed references to deprecated fields
* Remove sources dropdown from demographic data download


app/config/api_version
```diff
@@ -1 +1 @@
-a364195
\ No newline at end of file
+master
```
littledice: seeing as we're aiming at CI for the API :)

app/src/test/contrib/fetch/DemographicFetchTests.ts
```diff
@@ -38,54 +38,44 @@ describe("DemographicStore.fetchOneTimeToken", () => {
...
 
-        // Next, show that it does nothing if no source is selected
-        demographicActions.selectDataSet(dataSet.id);
-        demographicStore.fetchOneTimeToken().catch(doNothing);
-        expect(stub.called).to.equal(false, "Fetched token even though data set was not selected");
 
-        // Next, select a source as well and show that this makes the difference
-        demographicActions.selectSource(dataSet.sources[0]);
-        demographicStore.fetchOneTimeToken().catch(doNothing);
-        expect(stub.called).to.equal(true, "Did not fetch token even though data set was selected");
```
MartinEden: Previously this test showed that it didn't fetch until it was meant to, and then that it did fetch once you had selected all the options.

Now it just shows that it doesn't fetch if you don't select a data set. Shouldn't it also check that it does fetch when you have selected one?

PR #61, opened 2017-10-23
Package up integration tests as a Docker image so they can be run independently on this repository

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-633


app/docker/shared-build-env.dockerfile
```diff
@@ -0,0 +1,64 @@
+FROM node:7
```
MartinEden: This file has been copied with out changes from build.dockerfile.

app/scripts/add-test-accounts-for-integration-tests.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+set -ex
+here=$(dirname $0)
+
+$here/cli.sh add "Test User" test.user \
+    test@example.com password \
+    --if-not-exists
```
MartinEden: Idempotent, so multiple batches of portal integration tests can be run against the same database.

app/src/main/shared/Settings.ts
```diff
@@ -1,7 +1,8 @@
+import { settings as defaultSettings } from "./settings/default";
 const environment = require('environmentSettings');
 const appName = require('appName');
 
-export const settings: Settings = environment.settings;
+export const settings: Settings = Object.assign({}, defaultSettings, environment.settings);
```
MartinEden: Now we have a default set of settings, and the other files override, instead of every file having to specify all settings. About time!

app/config/api_version
```diff
@@ -1 +1 @@
-master
\ No newline at end of file
+b0c159c
```
littledice: can change this back to master now!

app/scripts/run-integration-tests-in-container.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+set -ex
+./scripts/add-test-accounts-for-integration-tests.sh
+./scripts/run-integration-tests.sh "$@"
```
littledice: can move env var setting to here from the Dockerfile

MartinEden:
Changes made as requested.

PR #64, opened 2017-10-26
Add reporting portal integration tests

littledice:


app/scripts/run-development-apis.sh
```diff
@@ -11,20 +11,10 @@ function on_interrupt() {
...
 }
 
-# Generate report test data
-docker pull $registry/orderly:$orderly_version
-docker run --rm \
-    --entrypoint create_orderly_demo.sh \
-    -v montagu_orderly_volume:/orderly \
-    $registry/orderly:$orderly_version \
-    /orderly
-
```
littledice: this moved to `run-apis.sh` as always needed when that script runs

app/scripts/run-apis.sh
```diff
@@ -15,10 +15,22 @@ docker run --rm \
...
+# Generate report test data
+docker pull $registry/orderly:$orderly_version
+docker run --rm \
+    --entrypoint create_orderly_demo.sh \
+    -v montagu_orderly_volume:/orderly \
+    $registry/orderly:$orderly_version \
+    /orderly
+
+docker exec montagu_reporting_api_1 sh -c 'cp /orderly/demo/. /orderly/ -r'
+
```
littledice: this is a workaround as the orderly task generates the 'demo'  folder we want as a subfolder of the orderly directory

app/scripts/run-apis.sh
```diff
@@ -15,10 +15,22 @@ docker run --rm \
     $registry/montagu-cert-tool:$cert_tool_version \
     gen-keypair /workspace
 
+docker volume prune -f
```
littledice: If the orderly volume hasn't been cleaned up for any reason, the below test generation task will fail

MartinEden: We shouldn't prune all volumes. It's fine for test/development tasks to stop and remove containers. But volumes are meant to be the persistent part. If the developer has anything on their machine in a volume that they actually want they could lose it by running this test.

Maybe you could just `docker volume rm` the appropriate volume?

littledice:
Ok, that makes sense. Let's merge this in as the first step (I've created a subtask which this PR can close (https://vimc.myjetbrains.com/youtrack/issue/VIMC-851) and will continue work on the larger task of getting full test coverage.

PR #68, opened 2017-10-31
Integration tests for reporting portal components

littledice:


app/src/integrationTests/ContribIntegrationTests.tsx
```diff
@@ -0,0 +1,361 @@
...
+                scenario: {
+                    id: scenarioId,
+                    description: "Yellow Fever scenario",
+                    disease: "yf",
+                    touchstones: [touchstoneId]
+                }
+            }
+        ]
+    };
+}
```
littledice: This file has also been renamed without changes

app/src/integrationTests/ReportIntegrationTests.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+                    const url = link.prop("href");
+                    return fetcher.fetchFromReportingApi(url)
+
+                });
+
+            checkPromise(done, promise, (response) => {
+                expect(response.status).to.equal(200)
+            });
+        });
```
MartinEden: In each of these file download tests there are some repeated lines that could be factored out:

```
const link = rendered.find(FileDownloadLink).first();
const url = link.prop("href");
return fetcher.fetchFromReportingApi(url)
```

and 

```
checkPromise(done, promise, (response) => {
    expect(response.status).to.equal(200)
});
```

littledice:
Ok, right number of tests now and less repeated code!

PR #72, opened 2017-11-03
Create dropdown of report versions

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-864

This took about 7 billion years longer than I had hoped, but this gives us a pattern for doing "redirects in response to a user action", which we didn't have before.

Changes to lots of files, but that's mostly just rippling through the fact that `PageProperties` now gives you access to the router. This is from the update to simple react router from [my PR in July](https://github.com/deadlyicon/simple-react-router/pull/7) to enable us to do this kind of stuff!


app/src/main/report/components/Versions/VersionDetails.tsx
```diff
@@ -8,34 +8,41 @@ import {ParameterList} from "../Parameters/ParameterList";
...
-    versionDetails: Version,
-    report: string
+interface PublicProps {
+    onChangeVersion: (version: string) => void;
+}
+
+export interface VersionProps extends RemoteContent, PublicProps {
+    versionDetails: Version;
+    report: string;
+    otherVersions: string[];
```
littledice: I think this shoudl be `allVersions` rather than `otherVersions` right?

app/src/main/report/components/Versions/VersionDetails.tsx
```diff
@@ -8,34 +8,41 @@ import {ParameterList} from "../Parameters/ParameterList";
...
+
+            onChangeVersion: props.onChangeVersion
         };
     }
 
-    renderContent(props: VersionProps) {
-        const p = props;
-        const url = `/reports/${p.report}/versions/${p.versionDetails.id}/all/`;
+    private renderTable(props: VersionProps) {
+        const url = `/reports/${props.report}/${props.versionDetails.id}/all/`;
```
littledice: one failing integration test - this url has changed!

app/src/main/report/components/Versions/ReportVersionSwitcher.tsx
```diff
@@ -0,0 +1,40 @@
...
+        versions.sort((a, b) => b.compareTo(a));
+        const items = versions.map(v => {
+            return <option value={v.id} key={v.id}>
+                {longTimestamp(v.timestamp)}
+            </option>
+        });
+        return <div className={"text-right mb-3 " + styles.versionSwitcher}>
+            Report version &nbsp;
+            <select onChange={this.onSelect} value={this.props.currentVersion}>
+                {items}
```
littledice: Cool, so you went with single line/vanilla html select. The only other place I think we use a select so far is on the download demographics page, where I've styled it with the bootstrap `form-control` class (see https://support.montagu.dide.ic.ac.uk:10443/contribution/Suraratdecha/responsibilities/201708test-2/demographics/). To keep styling consistent, we should either use that class here too, or remove from the demographics page...

Actually the bootstrap class adds a border-radius, which is a little inconsistent with the rest of the site. We should probably move to including Bootstrap's .scss files and compiling them using webpack rather than just including the pre-compiled css. Then we can make proper use of its variables, e.g. setting `$enable-rounded: false` will remove all rounded corners.

app/src/main/report/components/Versions/VersionDetails.tsx
MartinEden: Fixed

app/src/main/report/components/Versions/VersionDetails.tsx
MartinEden: Fixed

app/src/main/report/components/Versions/ReportVersionSwitcher.tsx
MartinEden: It turned out to be really complicated just as it was, so I thought I'd do vanilla select here, and we can always improve the look and feel later. I have added Bootstrap styling as discussed.

MartinEden:
Made changes as suggested. The error you're getting locally is probably because you need to run `npm install` to update simple-react-router. Simple react router handles passing the new router variable in.

littledice:
One failing unit test

PR #79, opened 2017-11-06
Use display name in report page title

MartinEden:
Merge https://github.com/vimc/montagu-webapps/pull/78 first.

Part of the https://vimc.myjetbrains.com/youtrack/issue/VIMC-868 work.

### DIFF
--- a/app/src/main/report/components/Reports/ReportPageTitle.tsx
+++ b/app/src/main/report/components/Reports/ReportPageTitle.tsx
@@ -14,11 +14,11 @@ export class ReportPageTitleComponent extends React.Component<Props, undefined> 
 
     static getPropsFromStores(): Props {
         const s = reportStore.getState();
-        const props = {
+        const props: Props = {
             name: s.currentReport,
             displayName: null
         };
-        const details = s.versionDetails[s.currentVersion];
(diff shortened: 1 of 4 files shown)


app/src/main/report/components/Reports/ReportPageTitle.tsx
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    static getPropsFromStores(): Props {
+        const s = reportStore.getState();
+        const props: Props = {
+            name: s.currentReport,
+            displayName: null
+        };
+        const details = s.versionDetails[s.currentReport];
+        if (details) {
```
littledice: would it ever be the case that the details have not  been fetched yet, and if so should we wait for them to be fetched before displaying?

MartinEden: Yes, it could happen: When you navigate to the page is when we kick off the API request to fetch the details. So we already have a loading spinner for the page content. I considered just making the title "Loading..." until the details have been fetched, but I figure if it does for some reason take a few seconds (due to slow Internet) it would be more informative for the user if we displayed the information we already had, and then switch to the better title when it arrived.

It most usage I think it will be very quick and the user won't really see the shorter title.

PR #81, opened 2017-11-08
https://vimc.myjetbrains.com/youtrack/issue/VIMC-397

MartinEden:
Version 2 of https://vimc.myjetbrains.com/youtrack/issue/VIMC-397

So the issue we had previously was that some breadcrumb URLs/names depended on store state. However, the breadcrumbs weren't necessarily getting setup after the store was ready. So the simplest thing to test was just that the breadcrumb creating code was getting called at the right time - i.e. after the page has finished fetching remote data. However, I then realised this was pretty meaningless without als checking that the data that was fetched matched up with the store state needed for the crumb. The solution to this was, as discussed, to add a kind of integration test to each page that fetches data into the stores.

The diff is obviously big here, because we backed out the previous PRs. Here's what has changed since the last time you reviewed this:

* We now have `NavigationTests.ts`, which is two tests that can be applied to each page. One checks normal navigation to the page and the other jumping straight to that URL as your entry point. The key thing they check is a) that the breadcrumb code does get invoked b) the url and page name are not null or undefined. They might be null or undefined if the breadcrumb creation is happening too early (before all required store state has been fetched) or because the breadcrumb depends on store state that is never fetched.
* Each of the individual page test files now invokes `addNavigationTests` to add these two tests, having first done any mocking of the API endpoints required.
* I've added a bunch more mocking infrastructure to support easily mocking out multiple endpoints at once.
* Finally, where the tests showed errors, I've gone through and made sure that `load()` in the page components is only calling `super.load` (and hence triggering breadcrumb creation) after sufficient data has been fetched.

So a fair bit of work, but we now have significantly more confidence that the user can browse to any URL and have everything work :)


littledice:
This is a lot of files! I know I've reviewed a lot of the code before, but could you split it up into smaller PRs again? Just don't feel I can effectively review so many file changes...

MartinEden:
I'm afraid I'm really not sure how to split it up. If it helps, of the 58 files changed, most of them are the pages that all need a breadcrumb now and the tests for those pages, and all of those files change in very predictable ways.

Maybe we can talk through it on Monday?

littledice:
Ok, lets talk through it on Monday.

PR #82, opened 2017-11-08
Visual indication of published status

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-865


MartinEden:
Good catch, thanks :)

PR #83, opened 2017-11-09
Request redirect back to app when POSTing estimates

littledice:
* set and add redirect url query parameter to request for one time link
* on redirection back to the app, parse returned query string and notify if an error has occurred

Pausing here to keep PR small, further issue opened to notify user of a successful upload:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-955


app/src/@types/simple-react-router.d.ts
```diff
@@ -23,7 +23,7 @@ declare module "simple-react-router" {
     export interface Location<T> {
         hash: string;
         params: T;
-        pathName: string;
+        pathname: string;
         query: any;
```
littledice: this was a bug in the type definition - the underlying code uses lowercase

app/src/main/shared/sources/JwtDecoder.ts
```diff
@@ -0,0 +1,8 @@
+const jwt_decode = require('jwt-decode');
+
+
+export const jwtDecoder = {
+    jwtDecode (token: string): any {
+        return jwt_decode(token)
+    }
+}
```
littledice: wrapped for easier spying/stubbing in tests

app/src/main/contrib/components/Responsibilities/BurdenEstimates/UploadBurdenEstimatesPage.tsx
```diff
@@ -1,29 +1,37 @@
...
+            processEncodedResultAndNotifyOnErrors<string>(queryStringAsObject());
+
             estimateTokenActions.clearUsedToken();
             modellingGroupActions.setCurrentGroup(this.props.location.params.groupId);
             responsibilityStore.fetchTouchstones().catch(doNothing).then(() => {
                 touchstoneActions.setCurrentTouchstone(this.props.location.params.touchstoneId);
                 responsibilityStore.fetchResponsibilities().catch(doNothing).then(() => {
                     responsibilityActions.setCurrentResponsibility(this.props.location.params.scenarioId);
+                    responsibilityActions.setRedirectPath(this.props.location.pathname);
                     responsibilityStore.fetchOneTimeEstimatesToken().catch(doNothing)
```
MartinEden: It might be a good idea to not directly expose `fetchOneTimeEstimatesToken` as part of the public signature of `responsibilityStore`, and instead expose as wrapper method that takes an optional redirect path. Then, the wrapper method can take care of calling `setRedirectPath`, and calling code is less error prone.

For an example of wrapping a fetch method like this, see `OneTimeTokenStoreInterface._fetchToken()`.

app/src/main/contrib/components/Responsibilities/BurdenEstimates/UploadBurdenEstimatesPage.tsx
```diff
@@ -1,29 +1,37 @@
...
     touchstoneId: string;
     scenarioId: string;
 }
 
-export class UploadBurdenEstimatesPage extends PageWithHeaderAndNav<LocationProps> {
+export class UploadBurdenEstimatesPage extends PageWithHeaderAndNav<UploadEstimatesProps> {
     componentDidMount() {
         setTimeout(() => {
+
+            processEncodedResultAndNotifyOnErrors<string>(queryStringAsObject());
```
MartinEden: Later, I think we should invoke this in `PageWithHeader`, so that all pages get this behaviour. For now though, I'm happy to keep this change less far reaching. Also, it would be better to leave changes to `PageWithHeader` until VIMC-397 is finished.

app/src/integrationTests/ContribIntegrationTests.tsx
```diff
@@ -235,8 +235,13 @@ class ContributionPortalIntegrationTests extends IntegrationTestSuite {
                 expect(payload).to.eql(JSON.parse(`{
                     ":group-id": "${groupId}",
                     ":touchstone-id": "${touchstoneId}",
-                    ":scenario-id": "${scenarioId}"
+                    ":scenario-id": "${scenarioId}"                    
                 }`));
+                const query = QueryString.parse(decoded.query);
+                expect(query).to.eql(JSON.parse(`{
+                    "redirectUrl": "http://localhost:5000/redirect/back"                   
+                }`
```
MartinEden: Why are you parsing this from a string, instead of just using an object literal?

littledice: I copied and pasted from the existing test tbh, haven't checked whether its necessary

MartinEden:
Again, I'll leave this for you to merge, since I can't see TeamCity.

PR #85, opened 2017-11-13
mount docker auth into container

littledice:


app/scripts/run-build.sh
```diff
@@ -1,5 +1,10 @@
 #!/usr/bin/env bash
+
+# This is the path for teamcity agents. If running locally, pass in your own docker config location
+docker_auth_path=${1:-/opt/teamcity-agent/.docker/config.json}
```
littledice: add local path to comment

PR #89, opened 2017-11-21
remove old templates

littledice:
And temporarily remove links


littledice:
That hasn't changed - I guess it never worked locally but will work when deployed. It has the link hard-coded as opposed to using the app settings, so will only work in the context of a full Montagu deploy. We can open a ticket to make this work locally but I don't think it needs to be fixed to close this ticket?

littledice:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-1068

PR #95, opened 2017-11-27
I1084 - Responsibility Overview Description and Guidance pages

georgearrowwood:
Responsibility Overview Page:
 * Changed Description text
 * Created separate sub component for Description
 * Created 2 help components named: model-inputs and model-otputs
 * Put links in Description to them

Other Changes:
 * Fixed breadcrumb links for help/touchstones
 * Added slackUrl to settings
 * Added .idea to .gitignore
 * Added scrollable anchor react module to application


app/src/main/contrib/components/Responsibilities/Overview/ResponsibilityOverviewPage.tsx
```diff
@@ -31,55 +32,21 @@ export class ResponsibilityOverviewPage extends ContribPageWithHeader<LocationPr
...
 
     title() {
-        return <ResponsibilityOverviewTitle />
+        return <ResponsibilityOverviewTitle />;
     }
 
     urlFragment(): string {
         const s = responsibilityStore.getState();
-        return `responsibilities/${s.currentTouchstone.id}/`;
+        return `responsibilities/${s.currentTouchstone ? s.currentTouchstone.id : ''}/`;
```
littledice: I don't think this is needed and in any case the path `/responsibilities/` would result in a 404 so there'd be no point directing users to that! This function will only get called either on this page after `super.load()` is called (by which time `s.currentTouchstone` will be set) or if this page was a parent for another page (which I don't think it is right?)

georgearrowwood:
made changes

PR #98, opened 2017-11-28
new source, actions, and updated tests

littledice:
This adds methods to get a onetime token for uploading model run parameters.


app/src/main/contrib/sources/ParametersTokenSource.ts
```diff
@@ -0,0 +1,26 @@
...
+    constructor() {
+        super();
+        this._fetchOneTimeParametersToken = () => this.doFetch(state => {
+            let queryString = "";
+            if (state.redirectPath && state.redirectPath.length > 0) {
+                queryString = "?redirectUrl=" + encodeURI(settings.montaguUrl() + state.redirectPath);
+            }
+            return `/modelling-groups/${state.currentModellingGroup.id}/model-run-parameters/`
+                + `${state.currentTouchstone.id}/get_onetime_link/${queryString}`;
+        }, {
```
MartinEden: Something weird with the indentation here

app/src/main/contrib/sources/ParametersTokenSource.ts
```diff
@@ -0,0 +1,26 @@
...
+
+    _fetchOneTimeParametersToken: () => SourceModel<string>;
+
+    constructor() {
+        super();
+        this._fetchOneTimeParametersToken = () => this.doFetch(state => {
+            let queryString = "";
+            if (state.redirectPath && state.redirectPath.length > 0) {
+                queryString = "?redirectUrl=" + encodeURI(settings.montaguUrl() + state.redirectPath);
+            }
```
MartinEden: The redirect path code is duplicated from elsewhere - could you refactor to this to be more DRY?

app/src/main/contrib/sources/ParametersTokenSource.ts
littledice: how would you format this? this is how my IDE wants to do it...

MartinEden: Oh, I realise now that I was misreading the code. But that makes me think even more that this isn't the clearest layout. How about [like this](https://github.com/vimc/montagu-webapps/commit/0d7833d7ed4b3d92df48b2a43c221ae47b6313e5)

PR #99, opened 2017-11-29
i978 - Display name is not rendered in title on report page

georgearrowwood:
to fix this i needed to fix the way version details are set in PageTitle component

### DIFF
--- a/app/src/test/report/components/Reports/ReportPageTitleTests.tsx
+++ b/app/src/test/report/components/Reports/ReportPageTitleTests.tsx
@@ -42,7 +42,7 @@ describe("ReportPageTitle", () => {
         expect(ReportPageTitleComponent.getPropsFromStores()).to.eql({
             name: "forecast",
             version: "v25",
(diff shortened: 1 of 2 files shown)


app/src/test/report/components/Reports/ReportPageTitleTests.tsx
```diff
@@ -42,7 +42,7 @@ describe("ReportPageTitle", () => {
         expect(ReportPageTitleComponent.getPropsFromStores()).to.eql({
             name: "forecast",
             version: "v25",
-            displayName: "Shipping Forecast"
+            displayName: null,//"Shipping Forecast"
```
MartinEden: The test was wrong in the first place, when I wrote it. It was setting the store up in a bogus state. So even though the component code I wrote was wrong, the test still passed. As part of this work, we should change the test so it sets up the store correctly. Then, the returned value will have the correct display name, I think, and this change will be unnecessary.

PR #105, opened 2017-11-30
i924: Report burden estimate set status to users

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-924

First review this: https://github.com/vimc/montagu-webapps/pull/104

As part of this I've factored out common logic between the responsibilities list and the estimate upload form. They now use a common component (`CurrentEstimateSetStatus`) to render info about the whether an estimate set has been uploaded, and what status it is in. This results in a slightly change to behaviour: We now repeat the notification that users can't upload any more estimates when the responsibility set is finished next to each scenario, as well as at the top of the responsibilities list, but I think this is probably a good thing.


app/package.json
```diff
@@ -4,6 +4,8 @@
   "description": "Modelling groups burden contribution portal",
   "main": "index.js",
   "scripts": {
+    "webpack": "webpack",
+    "watch_webpack": "webpack --watch",
```
littledice: `npm run webpack_watch` is actually more commands to type than `webpack --watch`, or am I missing something?

PR #107, opened 2017-11-30
i586 - on click download button it will be disabled for 5 seconds

georgearrowwood:
* to do this i needed to use local state on DownloadCoverageContentComponent
* to use local state i needed to create proper state interface for this component
* to do this i needed to change parent component RemoteContentComponent to be able to accept state interface
* this change made me change all components using RemoteContentComponent as parent
* now local state interface can be used if needed for all of them
* had to fix some tests

### DIFF
--- a/app/src/main/admin/components/ModellingGroups/List/ModellingGroupsList.tsx
+++ b/app/src/main/admin/components/ModellingGroups/List/ModellingGroupsList.tsx
@@ -10,7 +10,7 @@ interface ModellingGroupsProps extends RemoteContent {
     groups: ModellingGroup[]
 }
 
-export class ModellingGroupsListComponent extends RemoteContentComponent<ModellingGroupsProps, {}> {
+export class ModellingGroupsListComponent extends RemoteContentComponent<ModellingGroupsProps, undefined> {
     static getStores() {
         return [ groupStore ];
     }
--- a/app/src/main/admin/components/ModellingGroups/List/ModellingGroupsList.tsx
+++ b/app/src/main/admin/components/ModellingGroups/List/ModellingGroupsList.tsx
@@ -10,7 +10,7 @@ interface ModellingGroupsProps extends RemoteContent {
     groups: ModellingGroup[]
 }
 
-export class ModellingGroupsListComponent extends RemoteContentComponent<ModellingGroupsProps> {
+export class ModellingGroupsListComponent extends RemoteContentComponent<ModellingGroupsProps, {}> {
     static getStores() {
         return [ groupStore ];
     }
--- a/app/src/main/contrib/components/Responsibilities/Coverage/DownloadCoverageContent.tsx
+++ b/app/src/main/contrib/components/Responsibilities/Coverage/DownloadCoverageContent.tsx
@@ -17,27 +17,32 @@ const commonStyles = require("../../../../shared/styles/common.css");
 const styles = require("../Responsibilities.css");
 
 export interface DownloadCoverageComponentProps extends RemoteContent {
-    props: Props
-}
-
-interface Props extends HasFormatOption {
-    touchstone: Touchstone;
-    scenario: Scenario;
-    coverageSets: CoverageSet[];
-    coverageToken: string;
+    touchstone?: Touchstone;
+    scenario?: Scenario;
+    coverageSets?: CoverageSet[];
+    coverageToken?: string;
+    downloadButtonDisableTimeout?: number;
+    selectedFormat?: string;
 }
 
 interface DownloadState {
     downloadButtonEnabled: boolean;
 }
 
-export class DownloadCoverageContentComponent extends RemoteContentComponent<DownloadCoverageComponentProps, DownloadState> {
-    constructor() {
-        super();
+export class DownloadCoverageContentComponent
+    extends RemoteContentComponent<DownloadCoverageComponentProps, DownloadState>
+{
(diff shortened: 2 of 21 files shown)


MartinEden:
So far we have been strictly following stateless components, and using stores for state, as per the flux pattern. I think you have more experience with React then me and Alex, Juri, so I'm willing to be convinced, but in general I think that the pure flux pattern is much easier to test.

I don't see any tests here?

app/src/main/admin/components/ModellingGroups/List/ModellingGroupsList.tsx
```diff
@@ -10,7 +10,7 @@ interface ModellingGroupsProps extends RemoteContent {
     groups: ModellingGroup[]
 }
 
-export class ModellingGroupsListComponent extends RemoteContentComponent<ModellingGroupsProps> {
+export class ModellingGroupsListComponent extends RemoteContentComponent<ModellingGroupsProps, {}> {
```
MartinEden: In general, we have been using `undefined` rather than `{}` in these kind of cases.

littledice:
FWIW I am totally on board with using local state where appropriate (I have also used local state in the file upload form.) Just missing tests, but I think the approach makes sense.

littledice:
I think the top rated comment here makes a lot of sense, re when to use local vs store state! https://github.com/reactjs/redux/issues/1287

MartinEden:
Okay, fair enough, I'm convinced. In which case, Juri if you just use `undefined` instead of `{}` where I indicated and add some tests, this will be fine. :)

app/src/main/admin/components/ModellingGroups/List/ModellingGroupsList.tsx
georgearrowwood: * wrote 2 unit tests: 1 tests on download component and 1 on onetimebutton
* changed {} to undefined

app/src/main/contrib/components/Responsibilities/Coverage/DownloadCoverageContent.tsx
```diff
@@ -65,6 +76,19 @@ export class DownloadCoverageContentComponent extends RemoteContentComponent<Dow
         responsibilityStore.fetchOneTimeCoverageToken();
     }
 
+    onDownloadClicked() {
+        setTimeout(() => {
+            this.setState({
+                downloadButtonEnabled: false,
+            })
+        }, 50)
```
MartinEden: I may be wrong about this, but I'll just check: I think the 50 ms delay is unnecessary. If you leave the timing argument off I would expect the browser to finish resolving the user button action, and only when the stack was empty would this callback be invoked.

app/src/main/contrib/components/Responsibilities/Coverage/DownloadCoverageContent.tsx
```diff
@@ -65,6 +76,19 @@ export class DownloadCoverageContentComponent extends RemoteContentComponent<Dow
...
+        setTimeout(() => {
+            this.setState({
+                downloadButtonEnabled: true,
+            })
+        }, 5000)
```
MartinEden: Maybe more readable as

```
setTimeout(() => this.setState({
    downloadButtonEnabled: true
}), 5000);
```

But that's personal preference. Equally the other statement. Also, it might be nice to pull the 5000ms delay into a named static const?

app/src/main/shared/components/OneTimeButton.tsx
```diff
@@ -8,16 +8,17 @@ interface Props {
     token: string
     enabled?: boolean;
     refreshToken: () => void;
+    onClickOuterEvent?: () =>void;
```
MartinEden: I think this could just be called `onClick`. Better to have it easy to understand for components using the `OneTimeButton`. If you'd like to keep things unambiguous, maybe rename the internal method to something like `internalOnClickHandler`.

app/src/test/contrib/components/Responsibilities/DownloadCoverageComponentTests.tsx
```diff
@@ -58,6 +59,22 @@ describe("DownloadCoverageContentComponent", () => {
...
+        expect(component.state().downloadButtonEnabled).to.be.equal(true)
+        instance.onDownloadClicked();
+        setTimeout(() => {
+            expect(component.state().downloadButtonEnabled).to.be.equal(false)
+        },70);
+        setTimeout(() => {
+            expect(component.state().downloadButtonEnabled).to.be.equal(true)
+            done();
+        },5010);
+    });
```
MartinEden: Hmm, this makes this test take 5 seconds to run which is a bit long. Maybe parameterize the delay (via a React prop) so you can test this more rapidly?

app/src/main/contrib/components/Responsibilities/Coverage/DownloadCoverageContent.tsx
MartinEden: Although see below about idea to parameterize delay.

MartinEden:
Oh, I just remembered: There's a feature missing here. We also wanted to re-enable the button straight away if the user switches to a different demographic data set or option. If you like, that can be a separate PR.

georgearrowwood:
lets make this feature a separate PR

MartinEden:
Opened https://vimc.myjetbrains.com/youtrack/issue/VIMC-1119 for continuing work.

PR #108, opened 2017-11-30
fix bug with report resources in multiple nested subdirectories

littledice:


weshinsley:
So this fix solves x/y/file.csv - does it also solve the (decreasingly likely) x/y/z/file.csv - or should it go somewhere in the notes for writing reports that we should keep subfolder depth <=2?

weshinsley:
So this fix solves x/y/file.csv - does it also solve the (decreasingly likely) x/y/z/file.csv - or should it go somewhere in the notes for writing reports that we should keep subfolder depth <=2? -- actually, ignore this - I was looking at the tests. I think you've solved it for all cases.

app/src/main/shared/Helpers.ts
```diff
@@ -3,7 +3,7 @@ export function doNothing() {
 }
 
 export function encodeFilename(filename: string) {
-    return filename.replace("/", ":");
+    return filename.replace(/\//g, ":");
```
MartinEden: I find this regex a little unreadable. Perhaps it would be simpler to use the Regex constructor rather than a regex literal, like so: `new RegExp("/", "g")`

littledice: done :)

PR #110, opened 2017-12-04
remove unused props

littledice:
Noticed that `UploadForm` no longer uses `groupId` or `scenarioId` so have removed these from its props.


app/src/test/contrib/components/Responsibilities/Uploads/UploadFormTests.tsx
```diff
@@ -1,73 +1,68 @@
...
 
 const buttonStyles = require("../../../../../main/shared/styles/buttons.css");
-const messageStyles = require("../../../../../main/shared/styles/messages.css");
 
 describe('UploadForm', () => {
     let rendered: ShallowWrapper<any, any>;
     const sandbox = new Sandbox();
     before(() => mockFetcher(Promise.resolve(null)));
 
-    function setUpComponent(canUpload: boolean,
```
littledice: We don't actually need to set up the store for this component so this function was just obscuring the test logic.

PR #116, opened 2017-12-07
I1005 - Breadcrumb on report page should show version

georgearrowwood:
rely fully on initialise function in NavActions to create breadcrumb, making it canonical
removed navigate functionality
removed some tests

### DIFF
--- a/app/src/main/shared/stores/NavStore.ts
+++ b/app/src/main/shared/stores/NavStore.ts
@@ -24,7 +24,6 @@ class NavStore extends AbstractStore<NavState, AltJS.AltStore<NavState>> {
         super();
         this.bindListeners({
             handleInitialize: navActions.initialize,
-            handleNavigate: navActions.navigate
         });
     }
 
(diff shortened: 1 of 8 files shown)


app/src/test/shared/components/NavBarTests.tsx
```diff
@@ -17,18 +17,6 @@ describe("NavBar", () => {
         return shallow(<NavBarComponent crumbs={crumbs}/>).find("." + styles.chunk);
     };
 
-    it("can get props from store", () => {
-        navActions.navigate("a", "A");
```
littledice: I think this test should stay, just instead of using `navigate` to set up the store, use `alt.bootstrap` (see e.g. `ChooseGroupContentTests` for an example of bootstrapping an alt js store for testing.)

app/src/main/shared/stores/NavStore.ts
```diff
@@ -36,15 +35,6 @@ class NavStore extends AbstractStore<NavState, AltJS.AltStore<NavState>> {
         this.isInitialized = true;
```
littledice: this property can be deleted too now!

PR #119, opened 2017-12-11
Separate creation & population of burden estimates with reusable form component

littledice:
Reusable Form component uses the html5 form API to validate before submitting, and reports errors/success to the user. Note this commit: https://github.com/vimc/montagu-webapps/pull/119/commits/99cc09439303d451b3c04ef3bbbec6d8c3961ee3 shows the difference between using onchange handlers and passing down formatted data to the Form component, and using the native formData object along with a data formatting function passed down to the Form component.

The HTML5 form API is very sophisticated; most of our needs will likely be met using built in validation attributes, but you can also create arbitrary js rules if needed: https://developer.mozilla.org/en-US/docs/Learn/HTML/Forms/Form_validation


app/src/main/contrib/components/Responsibilities/BurdenEstimates/CreateBurdenEstimateSetForm.tsx
```diff
@@ -0,0 +1,92 @@
...
+        responsibilityStore.refreshResponsibilities()
+    }
+
+    onTypeChange(value: BurdenEstimateSetTypeCode) {
+        const newData = this.state.data;
+        newData.type.type = value;
+        this.setState({
+            data: newData
+        })
+    }
```
MartinEden: It might be better to use `CreateBurdenEstimateSet` directly as the state type. Then you can write state changes more succinctly as:

```
onDetailsChange(e: FormEvent<HTMLInputElement>) {
    this.setState((prevState) => {
        type: prevState.type.assign({
             details: e.target.value
        });
    })
}

onTypeChange(value: BurdenEstimateSetTypeCode) {
    this.setState((prevState) => {
        type: prevState.type.assign({
            type: value
        });
    })
}
```

Either way, isn't it better to use the prevState version of setState, or explicitly copy the old state object. It feels here like you are directly modifying the state object, which I think is not the normal React way of writing these functions.

app/src/main/contrib/components/Responsibilities/BurdenEstimates/CreateBurdenEstimateSetForm.tsx
```diff
@@ -0,0 +1,92 @@
...
+                    <div className="col">
+                        <label>Details of this run</label>
+                        <input type="text" className={"form-control"} name="details"
+                               onChange={this.onDetailsChange.bind(this)}/>
+                    </div>
+                </div>
+            </Form>
+        </div>;
+    }
+}
```
MartinEden: Nice! :)

app/src/main/shared/components/Alert.tsx
```diff
@@ -0,0 +1,55 @@
+import * as React from "react";
+
+interface AlertState {
+    show: boolean;
+    cssClass: string;
```
MartinEden: Can `cssClass` not just be derived from props in `render`? I don't think it needs to be in the state.

app/src/main/contrib/components/OptionSelector/OptionSelector.tsx
```diff
@@ -9,24 +9,31 @@ export interface OptionList {
...
-            { options }
-        </select>;
+        return this.props.required ? <select onChange={this.onChange} className={this.props.className} name={this.props.name} required>
+                {defaultOption}
+                {options}
+            </select> :
+            <select onChange={this.onChange} className={this.props.className} name={this.props.name}>
+                {defaultOption}
+                {options}
+            </select>;
```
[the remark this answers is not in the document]
littledice: oh good!

littledice:
V helpful comments, thanks! Made all your suggestions, plus got the integration test working and a few more unit tests.

littledice:
I haven't changed any existing forms, I'll add a ticket to the backlog to do so, so that we have consistency across the app.

littledice:
Lol, just checked Teamcity and this PR apparently adds 50 tests :open_mouth:

app/src/integrationTests/ContribIntegrationTests.tsx
```diff
@@ -272,6 +276,43 @@ class ContributionPortalIntegrationTests extends IntegrationTestSuite {
 
             });
         });
+
+        it("creates burden estimates set", (done: DoneCallback) => {
+
+            const rendered = mount(<CreateBurdenEstimateSetForm
+                touchstoneId={touchstoneId} groupId={groupId} scenarioId={scenarioId}/>);
+
+            const url = rendered.find(Form).prop("url");
```
MartinEden: One very minor request: Here you are just rendering the `CreateBurdenEstimateSetForm` component in order to get its URL. This confused me at first, as I was expecting `rendered` to be the subject under test. So a couple of possible changes: Either make the URL a static function on `CreateBurdenEstimateSetForm`, and then invoke it without having to render the component. Or keep the component as it is, but take these two lines out into a function called something like `getURLFromCreateBurdenEstimateSetForm`. In that case it would also be good to switch from `mount` to `shallow`, I think.

littledice: I've gone with the second suggestion, as I want to make sure I'm testing the url that is pulled from the actual props on the Form component.

PR #120, opened 2017-12-11
I920 - no css modular, css to sass

georgearrowwood:
* removed css modules functionality
* removed :locals from css files
* css classes as strings in html files
* changed .css extension to .scss
* added sass loader to webpack


littledice:
I can't exactly review this code but I have built locally and all looks good

PR #121, opened 2017-12-12
i1192 - Use bootstrap sass files

georgearrowwood:
* bootstrap is compiled now with webpack 
* created custom bootstrap bootstrap file (not all bootstrap modules are compiled, but can be activated if needed)
* created 3 mixins for buttons (color, border and outline) and one for fonts
* rewrited buttons style file using these mixins


littledice:
Bootstrap actually comes with some button mixins which make creating custom buttons v easy! Rather than creating all these custom mixins, we can just use the bootstrap ones:

No need to create any more mixins and our buttons are fully styled by:

```
@import "../../../../node_modules/bootstrap/scss/bootstrap.scss";

button, .button {
    font-weight: bold;
  @include button-variant(rgb(0, 173, 239), rgb(200, 200, 200));
    @include button-size(6px, 20px, 12pt, 24px, 0)
}

button[type=submit], .submit{
 @include button-variant(#5cb85c,  #4cae4c)
```

app/src/main/shared/styles/bootstrap.scss
```diff
@@ -0,0 +1,42 @@
+/*!
+ * Bootstrap v4.0.0-beta.2 (https://getbootstrap.com)
+ * Copyright 2011-2017 The Bootstrap Authors
+ * Copyright 2011-2017 Twitter, Inc.
+ * Licensed under MIT (https://github.com/twbs/bootstrap/blob/master/LICENSE)
+ */
+
+@import "../../../../node_modules/bootstrap/scss/functions";
+@import "../../../../node_modules/bootstrap/scss/variables";
```
littledice: At some point we'll want to replace this variables file with our own custom variables, although this is fine for now

app/src/main/shared/styles/mixins.scss
```diff
@@ -0,0 +1,39 @@
+@import "bootstrap";
```
littledice: no need for any of these, just use the `button-variant` mixins as in my comment

app/src/main/shared/styles/buttons.scss
```diff
@@ -1,28 +1,18 @@
...
+
 button, .button {
-    display: inline-block;
-    border: none;
-    background-color: rgb(0, 173, 239);
-    color: white;
-    font-family: inherit;
-    font-size: 12pt;
+    @include button-variant(rgb(0, 173, 239), rgb(200, 200, 200));
+    @include button-size(5px, 20px, 12px, 24px,  0);
```
littledice: Can we actually make the font size and line-height 1em and 2em? so `@include button-size(5px, 20px, 1em, 2em,  0);`

app/src/main/shared/styles/buttons.scss
```diff
@@ -1,28 +1,18 @@
...
-button:hover, .button:hover {
-    outline-width: 2px;
-}
-
-button:active {
-    outline-style: inset;
+    @include button-variant(#5cb85c, #4cae4c);
+    @include button-size(5px, 20px, 12pt, 24px,  0);
+    outline: #4cae4c outset 1px;
+    font-weight: bold;
```
littledice: Ditch this outline property and you don't need font-weight bold as `.submit` should always be applied to a button. So really should be `button.submit, .button.submit`

app/src/main/shared/styles/buttons.scss
```diff
@@ -31,30 +21,26 @@ button:disabled, .button.disabled {
...
-    color: #333;
-    background-color: #d4d4d4;
-    border-color: #8c8c8c;
-}
-
-button:disabled:hover, .button.disabled:hover {
-    outline-width: 1px;
+    @include button-variant(#d4d4d4, #333);
+    @include button-size(5px, 20px, 12pt, 24px,  0);
+    font-weight: bold;
```
littledice: again, don't need font-weight bold as already defined in `button, .button`

app/webpack.config.js
```diff
@@ -46,8 +46,7 @@ function makePortalConfig(name, urlPrefix) {
             new CopyWebpackPlugin([
                 { from: 'node_modules/react/dist/react.js', to: 'react.js' },
                 { from: 'node_modules/react-dom/dist/react-dom.js', to: 'react-dom.js' },
-                { from: 'node_modules/bootstrap/dist/css/bootstrap.min.css', to: 'bootstrap.min.css' },
-                { from: 'node_modules/bootstrap/dist/css/bootstrap.min.css.map', to: 'bootstrap.min.css.map' }
+                // { from: 'node_modules/sass-bootstrap/sass', to: 'bootstrap.min.css' },
```
littledice: can we delete this commented line?

app/src/main/shared/styles/buttons.scss
```diff
@@ -31,30 +21,26 @@ button:disabled, .button.disabled {
...
     box-shadow: none;
     opacity: .65;
-    color: #333;
-    background-color: #d4d4d4;
-    border-color: #8c8c8c;
-}
-
-button:disabled:hover, .button.disabled:hover {
-    outline-width: 1px;
+    @include button-variant(#d4d4d4, #333);
```
littledice: this is going to give it a dark border, just make the border color the same as the background, `#d4d4d4`

littledice:
Nearly there! Just a few tweaks needed.

app/src/main/shared/styles/buttons.scss
```diff
@@ -1,28 +1,15 @@
...
-    outline-color: #4cae4c;
-}
-
-button:hover, .button:hover {
-    outline-width: 2px;
-}
-
-button:active {
-    outline-style: inset;
+    @include button-variant(#5cb85c, #4cae4c);
```
littledice: one last thing! I do think lets change this rule from just `.submit` to `button.submit, `.button.submit`

PR #128, opened 2017-12-14
Use bootstrap table styling.

MartinEden:
Please take a look at this. I think in some cases this is a little less clear, as there is no bottom border to the default bootstrap style for tables. Maybe we should add one?

Part of the problem is that we still use tables for data that is not really tabular. The division between table and not-table is clearer when the table has more than a couple of rows.


app/src/main/contrib/components/Responsibilities/Coverage/DownloadCoverageContent.tsx
```diff
@@ -112,7 +112,8 @@ export class DownloadCoverageContentComponent
             </div>
             <div className="row mt-4">
                 <div className="col-12 col-md-6">
-                    <table className="options">
+                    <div className="smallTitle">Choose format</div>
```
MartinEden: Was blending into table above, so I added a new header to break things up.

littledice:
I prefer our current styling, with borders. Happily there is a Bootstrap class for this, `table-bordered`! The default Bootstrap border color is that light grey, but if we prefer our darker border, this is a good use case for introducing a custom variables file. If we copy the default bootstrap `variables.scss` file and replace `$table-border-color` with our preferred color, then compile with that instead, it should work nicely.

MartinEden:
I've added table-bordered, but it makes me a little unhappy to have so much repetition. Would it make sense to factor this out somehow. e.g: Create a global const string that is those three classes? Or create a React component called `Table` that just passes everything through to a native `table`, but adds the three classes? Neither approach seems great...

I'm not sure how to do the variables.scss thing - I can't see a file like that anywhere in our hierarchy. Do you mean copying from some source distribution of Bootstrap?

littledice:
Ok, here's what I mean with the variables, plus one way of reducing repetition with css:
https://github.com/vimc/montagu-webapps/pull/132/files

.gitignore
```diff
@@ -32,5 +34,3 @@ nohup\.out
 react-prototype/dist/*
 react-prototype/node_modules/*
 react-prototype/npm-debug\.log
-.idea/*
-!app/\.idea/codeStyleSettings.xml
```
MartinEden: gitignore changes unrelated to this ticket, but my IDE was throwing out a few extra files I didn't want to commit. I noticed while I was doing it that we had a couple of pointless rules:

* `.idea/*`: There is no .idea folder in the root, and we don't want to exclude the whole .idea folder, just individual files
* `!app/\.idea/codeStyleSettings.xml`: since we're not excluding the whole folder, there's no need to re-include this one file.

PR #132, opened 2017-12-14
I1212 alex suggestions

littledice:


app/src/main/shared/styles/variables.scss
```diff
@@ -0,0 +1,828 @@
+// Variables
```
littledice: This file is copied from bootstrap source files. I have changed the table-border-color var to a darker grey...

PR #142, opened 2017-12-22
i562 - Only show touchstones where selected modelling group has a responsibility

georgearrowwood:
Only show touchstones where selected modelling group has a responsibility

### DIFF
--- a/app/src/main/contrib/stores/ResponsibilityStore.ts
+++ b/app/src/main/contrib/stores/ResponsibilityStore.ts
@@ -42,7 +42,7 @@ export interface ResponsibilityState extends RemoteContent, HasFormatState {
 interface ResponsibilityStoreInterface extends AltJS.AltStore<ResponsibilityState> {
     fetchResponsibilities(): Promise<Responsibilities>;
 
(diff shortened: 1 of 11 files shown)


app/src/main/contrib/stores/ResponsibilityStore.ts
```diff
@@ -42,7 +42,7 @@ export interface ResponsibilityState extends RemoteContent, HasFormatState {
 interface ResponsibilityStoreInterface extends AltJS.AltStore<ResponsibilityState> {
     fetchResponsibilities(): Promise<Responsibilities>;
 
-    fetchTouchstones(): Promise<Touchstone[]>;
+    fetchTouchstones(groupId?: string): Promise<Touchstone[]>;
```
MartinEden: I don't think this should have changed - it refers to the method on the `TouchstoneSource`, which still takes no parameters.

MartinEden:
Once that unnecessary parameter is removed (assuming you agree @georgearrowwood) you can go straight ahead and merge this PR and mark the ticket as "Ready to deploy".

Merry Christmas!

PR #148, opened 2018-01-03
cleanup on error

littledice:
This seems to work, am I using this right @richfitz?


richfitz:
yeah, though you can do it even more simply by letting the cleanup function *always* do the cleanup (not explicitly cleaning)

```
function cleanup {
    set +e
    $here/stop-apis.sh
}
trap cleanup EXIT

$here/run-integration-tests.sh "$@"
result=$?

exit $result
```

PR #149, opened 2018-01-03
I1215

littledice:


littledice:
All logic here duplicated in other smaller branches

PR #153, opened 2018-01-05
bring in reactstrap

littledice:
* replaced our alert with reactstrap alerts
* updated tests
* add encoding type to form and update form submit method, so that if data is null it submits the form data


app/src/main/shared/components/Form.tsx
```diff
@@ -76,22 +79,33 @@ export class Form extends React.Component<FormProps, FormState> {
...
+            this.submitForm(form)
         }
     }
 
+    onChange(){
+        this.setState({
+            errors: [],
+            hasSuccess: false
+        });
+    }
```
littledice: doing this on the form change event so that alerts get cleared when the form inputs change

app/src/main/shared/components/Form.tsx
```diff
@@ -33,13 +33,15 @@ export class Form extends React.Component<FormProps, FormState> {
         }
     }
 
-    submitForm() {
+    submitForm(form: HTMLFormElement) {
 
         const self = this;
```
georgearrowwood: arrow functions should pass `this`, so `self` is not really needed

app/src/main/shared/components/Form.tsx
```diff
@@ -76,22 +79,33 @@ export class Form extends React.Component<FormProps, FormState> {
...
+
     render() {
 
         const hasError = this.state.errors.length > 0;
-        const alertMessage = hasError ? this.state.errors[0].message : this.props.successMessage;
 
         return <div>
-            <form className={`mt-4 ${this.state.validated ? "was-validated" : ""}`} onSubmit={this.onSubmit.bind(this)}
+            <form encType="multipart/form-data" className={this.state.validated ? "was-validated" : ""}
+                  onSubmit={this.onSubmit.bind(this)}
```
georgearrowwood: would move binds to constructor, e.g.
this.onSubmit = this.onSubmit.bind(this)

PR #158, opened 2018-01-08
I1179 - link to model run params page

littledice:
* added 'parameters' section to responsibilities overview page
* made guidance link on model run params page scroll to relevant section


MartinEden:
As discussed, please restore `control` styling to the disease filter dropdown. Otherwise, looks fine :)

PR #163, opened 2018-01-15
I1296 - call set shiny cookie endpoint

littledice:
But first merge https://github.com/vimc/montagu-api/pull/186 and point this branch back to master


littledice:
Yep, the js can't read the cookie. We could return the `allowed_shiny` value from the endpoint though (at the moment we just set the cookie header and return "OK") ?

MartinEden:
No, that's fine. We should be able to debug (if necessary) just by looking at the browser's cookie state separately.

littledice:
That's right, slightly more steps to get the claims but not too hard

PR #164, opened 2018-01-16
I1338 - bug, still using deprecated endpoint

littledice:


app/src/test/contrib/stores/ResponsibilityStoreTests.ts
```diff
@@ -294,18 +294,6 @@ describe("ResponsibilityStore", () => {
         expect(state.estimatesOneTimeToken).to.equal(null);
     });
 
-    it("estimateTokenActions.clearUsedToken sets token to null", () => {
```
littledice: not to do with current changes, just noticed a duplicate test

app/src/main/contrib/actions/EstimateActions.ts
```diff
@@ -1,14 +1,21 @@
-import { FetchActions, FetchActionsInterface } from "../../shared/actions/FetchActions";
-import { alt } from "../../shared/alt";
+import {FetchActions, FetchActionsInterface} from "../../shared/actions/FetchActions";
+import {alt} from "../../shared/alt";
 
 interface Actions extends FetchActionsInterface<string> {
     clearUsedToken(): boolean;
+
+    setRedirectPath(path: string): string;
```
littledice: This has to be separated into it's own action as is called at the page level, whereas the token needs to be fetched by the form, once the burden estimate set has been created and we have an id

app/src/integrationTests/ContribIntegrationTests.tsx
```diff
@@ -299,12 +309,26 @@ class ContributionPortalIntegrationTests extends IntegrationTestSuite {
...
+                .then((responsibilityIds) => {
+                    return addModel(db)
+                        .then((modelVersionId) => {
+                            return addBurdenEstimateSet(db, responsibilityIds.responsibility, modelVersionId)
+                                .then((setId) => {
+                                    return updateCurrentBurdenEstimateSet(db, responsibilityIds.responsibility, setId)
+                                })
+                        })
+                })
+        }
```
MartinEden: Maybe more readable like:
```
return addResponsibilities(db)
    .then(responsibilityIds => addModel(db))
    .then(modelVersionId => addBurdenEstimateSet(db, responsibilityIds.responsibility, modelVersionId))
    .then(setId => updateCurrentBurdenEstimateSet(db, responsibilityIds.responsibility, setId));
```

PR #170, opened 2018-01-24
i355: Reduce duplication by moving shared logic into parent pages

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-355


MartinEden:
So there's an issue here I haven't figured out how to resolve elegantly. We need to use the URL parameters to load data. When we delegate to the parent, the parent needs to be able to get at those URL parameters. Each child uses the same URL as the parent as a root and then extends it, adding more URL parameters.

So clearly there should be an inheritance chain: PageC is the child of PageB, which is the child of PageA. And PageC's URL props object should be a subclass (with additional properties) of PageB's URL props object, which should be a subclass of PageA's URL props object.

But darned if I can figure out how to make that explicit in the type system. I've cheated and made `IPageWithParent.load` take an `any` object. Any better suggestions?

littledice:
I can't think of any clever way of doing this unfortunately, but as discussed we think it's fairly safe as is.

PR #171, opened 2018-01-25
i652: If user is a member of non-existent modelling group, the modelling group portal will not load 

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-652

Turns out this bug has already been fixed in the past. This PR adds a test that proves the correct behaviour is present.

### DIFF
--- a/app/src/test/contrib/stores/ContribAuthStoreTests.ts
+++ b/app/src/test/contrib/stores/ContribAuthStoreTests.ts
@@ -55,6 +55,19 @@ class ContribAuthStoreTests extends AuthStoreBaseTests<ContribAuthState, Contrib
             expect(contribAuthStore.getState().modellingGroups).to.eql([ group ]);
         });
 
+        it("if the user is a member of a non-existent group, this is ignored", () => {
+            const realGroup = mockModellingGroup({ id: "test.group" });
+
+            alt.bootstrap(JSON.stringify({
+                ContribAuthStore: {
+                    modellingGroupIds: [ realGroup.id, "fake.group" ]
+                }
+            }));
+
+            modellingGroupActions.updateGroups([ realGroup ]);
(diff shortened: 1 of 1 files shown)


app/src/test/contrib/stores/ContribAuthStoreTests.ts
```diff
@@ -55,6 +55,19 @@ class ContribAuthStoreTests extends AuthStoreBaseTests<ContribAuthState, Contrib
...
+        it("if the user is a member of a non-existent group, this is ignored", () => {
+            const realGroup = mockModellingGroup({ id: "test.group" });
+
+            alt.bootstrap(JSON.stringify({
+                ContribAuthStore: {
+                    modellingGroupIds: [ realGroup.id, "fake.group" ]
+                }
+            }));
+
+            modellingGroupActions.updateGroups([ realGroup ]);
```
littledice: I'm not sure I understand this - why is `updateGroups` called with only the real group?

MartinEden: So there are two sources of information for the auth store:

1. The auth token, which contains (amongst other things) group *membership" - just the IDs of the groups
1. The `/modelling-groups/` endpoint, which returns full info (i.e. id, name, description) of all groups in the database.

The contrib store combines these two to get an array that contains full info on just the groups the user is a member of.

So here, the `updateGroups` action is mimicking what happens when we fetch all modelling groups from the API. Because the fake group only exists in a role mapping, but not actually in the database, this group would not be present in this list returned from the API.

littledice: Cool, I suspected something along these lines but struggled to follow the chain of events in the code. But this makes sense.

MartinEden: @littledice Sorry it's not clearer in the code!

PR #176, opened 2018-02-05
i1420: Alphabetize user list and dropdown

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-1420


app/docker/shared-build-env.dockerfile
```diff
@@ -52,7 +52,7 @@ RUN ./src/webmodels/gradlew
 
 # Install Node dependencies
 COPY package.json .
-RUN npm install
```
MartinEden: This is unrelated, but it makes the TeamCity build logs *massively* less verbose. Like, thousands of lines less.

PR #178, opened 2018-02-07
i1422: Check file extension is .csv before submitting estimate form 

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-1422

First review #177.

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -5,7 +5,7 @@
   "main": "index.js",
   "scripts": {
     "webpack": "webpack",
-    "watch_webpack": "webpack --watch",
+    "watch": "webpack --watch",
     "generate-models": "cd ./src/webmodels && ./gradlew generateTypeDefinitions",
     "//": "This is a comment: To run just one test, or a subset, use run-test.sh, e.g. ./run-test src/test/components/ErrorLogTests.tsx",
     "test": "mocha-webpack --webpack-config webpack-test.config.js \"src/test/**/*\"",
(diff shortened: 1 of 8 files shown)


app/package.json
```diff
@@ -5,7 +5,7 @@
   "main": "index.js",
   "scripts": {
     "webpack": "webpack",
-    "watch_webpack": "webpack --watch",
+    "watch": "webpack --watch",
```
MartinEden: Just an opportunistic change: I don't think anyone else uses this script other than me; just wanted to shorten the command I have to type.

app/src/test/Sandbox.ts
```diff
@@ -31,6 +31,7 @@ export class Sandbox {
         });
         this.mounted = [];
         alt.recycle();
+        resetFetcher();
```
MartinEden: I was getting some spurious test failures where other tests had left the fetcher in a weird state. The problem with global state! This helps, as any test that uses the sandbox (which is most of them) will then clear up after itself.

app/src/test/shared/components/UploadFormTests.tsx
```diff
@@ -5,11 +5,16 @@ import {Sandbox} from "../../Sandbox";
...
+import {CustomValidationResult} from "../../../main/shared/validation/FileValidationHelpers";
+import {mockFetcher, resetFetcher} from "../../mocks/mockRemote";
+import fetcher from "../../../main/shared/sources/Fetcher";
+import {ReportingFetcher} from "../../../main/report/sources/ReportingFetcher";
 
 describe('UploadForm', () => {
     let rendered: ShallowWrapper<any, any>;
     const sandbox = new Sandbox();
 
+    before(() => resetFetcher());
```
MartinEden: Apparently which ever test is interfering with these tests wasn't using sandbox, so I also ended up having to put this here. If we ever track down the test that's polluting the state, this can be removed again. In the mean time, there's no harm in this getting run more often that is needed.

PR #182, opened 2018-02-13
Couple of suggestions for typings and tests

littledice:
Added a few typings to components. Also a suggested pattern for testing `mapStateToProps` and `mapDispatchToProps` - define as static functions on the component itself, and then can test them in isolation.

### DIFF
--- a/app/src/main/contrib/components/ChooseGroup/ChooseGroupContent.tsx
+++ b/app/src/main/contrib/components/ChooseGroup/ChooseGroupContent.tsx
@@ -19,7 +19,7 @@ export interface ChooseGroupProps {
     ready: boolean;
 }
 
-export class ChooseGroupContentComponent extends React.Component<ChooseGroupProps, any> {
+export class ChooseGroupContentComponent extends React.Component<ChooseGroupProps, undefined> {
     componentDidMount() {
         this.props.getGroups()
     }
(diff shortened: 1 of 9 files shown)


app/src/main/contrib/components/ChooseGroup/ChooseGroupContent.tsx
```diff
@@ -6,19 +6,22 @@ import { GroupList } from "./GroupList";
...
+import { Dispatch } from "redux";
 
 import "../../../shared/styles/common.scss";
+import {ContribAppState} from "../../reducers/contribReducers";
+import {doNothing} from "../../../shared/Helpers";
 
 export interface ChooseGroupProps {
     groups: ModellingGroup[];
-    dispatch: any;
+    getGroups: () => Promise<void>;
```
georgearrowwood: Action get groups is not returning promise.

app/src/main/shared/reducers/GlobalState.ts
```diff
@@ -1,5 +1,6 @@
 import { AdminAppState } from "../../admin/reducers/adminReducers";
 import { ContribAppState } from "../../contrib/reducers/contribReducers";
 import { ReportAppState } from "../../report/reducers/reportReducers";
+import {AuthState} from "./authReducer";
```
georgearrowwood: not needed import

app/src/main/contrib/components/ChooseGroup/ChooseGroupContent.tsx
littledice: Typescript thinks it is? Because it's async?

PR #184, opened 2018-02-16
I1522 - Switch Report List to Redux

georgearrowwood:
Switch Report List to Redux


app/src/test/report/components/Reports/ReportListComponentTests.tsx
```diff
@@ -1,30 +1,21 @@
...
-                ready: true,
-                reports: reports
-            }
-        }));
-
-        expect(ReportListComponent.getPropsFromStores()).to.eql({
-            ready: true,
-            reports: reports
-        });
+        const reports = [ mockReport(), mockReport() ];
```
littledice: The name of this test should probably be changed to just `renders reports`? And then we need to test the `map...` functions that actually map the props from the global state.

app/src/main/report/actions/reportsActions.ts
```diff
@@ -0,0 +1,23 @@
...
+    getReports() {
+        return async (dispatch: Dispatch<ReportsActionsTypes>, getState: () => GlobalState) => {
+            const reports: Report[] = await (new ReportsService(dispatch, getState)).getAllReports();
+            dispatch({
+                type: ReportTypeKeys.REPORTS_FETCHED,
+                data: reports
+            } as ReportsFetched );
+        }
+    },
+
```
littledice: So can the `getReports` method from the old report source be deleted now? And the old action & listener on the old store?

app/src/main/report/components/MainMenu/MainMenu.tsx
```diff
@@ -1,19 +1,12 @@
...
 import {IPageWithParent} from "../../../shared/models/Breadcrumb";
 import { Page } from "../../../shared/components/PageWithHeader/Page";
 
-const menuStyles = require("./MainMenu.css");
+import "./MainMenu.scss";
 
 export class MainMenu extends ReportingPageWithHeader<undefined> {
-    load(props: undefined) {
-        return this.loadParent(props).then(() => {
-            return reportStore.fetchReports();
```
littledice: So is this just not needed now because it happens in the child component? And bc this page doesn't have a parent?

app/src/main/report/actions/reportsActions.ts
littledice: And the old `ReportStoreState` can now have `reports` removed...

app/src/main/report/components/MainMenu/MainMenu.tsx
georgearrowwood: there is no need to load list of reports on details page

littledice: Sorry I meant the `ReportList` component, child in the compositional sense!

app/src/main/report/stores/ReportStore.ts
```diff
@@ -50,7 +46,7 @@ class ReportStore
 
     initialState(): ReportStoreState {
         return {
-            reports: [],
+            // reports: [],
```
littledice: can delete commented code

PR #185, opened 2018-02-20
Re-organise stylesheets

littledice:
This is a preliminary refactor to the reporting app re-style work. I've re-organised our styles to allow for app level variables (e.g. header colours, link colours)

* all component level styles moved into styles folder (shared or app specific)
* all scss import statements removed from individual components
* each app has a master `style.scss` which sets custom variables, imports bootstrap and our common styles
* all app styles imported just once in each `index.tsx` file
* colour of reporting app navbar changed to lighter gray using app level variable


app/src/main/contrib/components/ChooseAction/ChooseTouchstone.scss
```diff
@@ -1,10 +0,0 @@
-.choice {
```
littledice: These styles are no longer needed as bootstrap has equivalents

app/src/main/contrib/components/ChooseGroup/ChooseGroup.scss
```diff
@@ -1,10 +0,0 @@
-:local(.choice) {
-    margin-bottom: 10px;
```
littledice: similarly, these duplicate existing bootstrap styles so aren't needed

app/src/main/report/components/MainMenu/MainMenu.scss
```diff
@@ -1,13 +0,0 @@
-.menu {
```
littledice: these styles are obsolete, bootstrap has similar styles (and we're redesigning this page anyway)

app/src/main/shared/styles/bootstrap.scss
```diff
@@ -18,13 +16,13 @@
 @import "../../../../node_modules/bootstrap/scss/tables";
 @import "../../../../node_modules/bootstrap/scss/forms";
 @import "../../../../node_modules/bootstrap/scss/buttons";
-// @import "transitions";
+@import "../../../../node_modules/bootstrap/scss/transitions";
```
littledice: added some imports that I need for the report page redesign

app/src/main/shared/styles/fonts.scss
```diff
@@ -1,9 +1,9 @@
 @font-face {
     font-family: header;
-    src: url("../resources/PetitaBold.ttf");
+    src: url("../shared/resources/PetitaBold.ttf");
 }
 
 @font-face {
     font-family: body;
-    src: url("../resources/Cloud-Light.otf");
+    src: url("../shared/resources/Cloud-Light.otf");
```
littledice: because this file gets imported into each app's `style.scss` file, the resource path needs to be relative to an individual app's directory

app/src/main/shared/styles/messages.scss
```diff
@@ -1,13 +0,0 @@
-.message {
-    font-size: 16pt;
```
littledice: none of these styles were being used any more

app/src/main/shared/styles/navbar.scss
```diff
@@ -1,5 +1,5 @@
 .montagu-navbar {
-    background-color: rgb(200, 200, 200);
+    background-color: $montagu-nav-bg-color;
```
littledice: allows for app level customisation

PR #186, opened 2018-02-20
I1523 - report details page

georgearrowwood:


app/src/test/report/components/Reports/ReportPageTests.tsx
```diff
@@ -1,85 +1,59 @@
...
-            const promise = page.changeVersion("versionname")
+        sandbox.setStub(ReportPageComponent.prototype, "loadVersion");
+        const location = mockLocation<ReportPageProps>({report: "reportname", version: "oldVersion"});
+        const page = new ReportPageComponent({location: location, router: router});
+        page.changeVersion("versionname")
+        setTimeout(() => {
             expect(redirectTo.called).to.equal(true, "Expected redirectTo to be called");
             expect(redirectTo.calledWith("/reportname/versionname/"));
-            return promise;
+            done();
```
littledice: So the previous version of this test also tested whether the actions (set current report, fetch version details etc) were triggered - would be good to keep that in the new test.

app/src/main/report/actions/reportsActions.ts
```diff
@@ -17,7 +20,38 @@ export const reportsActions = {
...
     },
 
+    setCurrentReport(report: string) {
+        return {
+            type: ReportTypeKeys.SET_CURRENT_REPORT,
+            data: report
+        }
+    },
+
+    setCurrentVersion(version: string) {
```
littledice: I'm wondering if we can do away with `currentVersion` altogether now - the reason we needed it before is because of the way we cached version details in the store - we had a lookup from version ids to version details, and when `currentVersion` changed we looked up the current details accordingly. But since we're moving away from that, I think we can always just get the current version id from `versionDetails.id`...

app/src/main/report/components/Reports/ReportPageTitle.tsx
```diff
@@ -1,41 +1,27 @@
...
-        const title = this.props.displayName || this.props.name;
-        return <span>
-            <div>{title}</div>
-            <div className="titleAddition">
-                Version: {this.props.version}
-            </div>
-        </span>;
+export const mapStateToProps = (state: ReportAppState): ReportPageTitleProps => {
+    return {
+        version: state.reports.currentVersion,
```
littledice: So for example here we could just use `state.reports.versionDetails.id`

app/src/main/report/components/Reports/ReportPage.tsx
```diff
@@ -1,42 +1,51 @@
...
-            reportActions.setCurrentVersion(props.version);
-            return reportStore.fetchVersionDetails();
+    componentDidMount() {
+        this.loadVersion();
+    }
+
+    changeVersion(version: string): any {
+        this.redirectToVersion(version);
+        setTimeout(()=> {
+            this.loadVersion();
```
littledice: I'm curious, why does this need to be wrapped in a timeout?

PR #187, opened 2018-02-20
I1518 sidebar

littledice:
Adds sidebar to report page. At the moment it doesn't do anything - the `Downloads` tab is marked as selected, and the 2 other tabs are disabled with tooltips to let the user know they're 'coming soon'.

Collapses to a burger menu on mobile.

### DIFF
--- a/app/src/main/report/components/Reports/ReportDetails.tsx
+++ b/app/src/main/report/components/Reports/ReportDetails.tsx
@@ -34,8 +34,8 @@ export class ReportDetailsComponent extends RemoteContentComponent<ReportDetails
             report: s.currentReport,
             allVersions: s.versions[s.currentReport],
             ready: s.ready
-                && s.versions[s.currentReport] !== undefined
-                && s.versionDetails[s.currentVersion] != null,
+            && s.versions[s.currentReport] !== undefined
+            && s.versionDetails[s.currentVersion] != null,
 
             onChangeVersion: props.onChangeVersion
         };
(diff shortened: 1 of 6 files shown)


app/src/main/report/components/Reports/ReportDetails.tsx
```diff
@@ -44,13 +44,15 @@ export class ReportDetailsComponent extends RemoteContentComponent<ReportDetails
     renderContent(props: ReportDetailsProps) {
         const version = props.versionDetails.id;
         return <div>
-           <DraftStamp published={props.versionDetails.published} />
+            <h1 className={"h2"}>{props.versionDetails.displayname || props.versionDetails.name}</h1>
+            <p className={"small text-muted"}>{props.versionDetails.id}</p>
```
littledice: The report title has moved down as we're no longer using the `Page` component with integrated title

app/src/main/report/components/Reports/Sidebar.tsx
```diff
@@ -0,0 +1,51 @@
...
+        this.setState({
+            isOpen: !this.state.isOpen
+        });
+    }
+
+    render() {
+        return <div className={"sidebar pb-4 pb-md-0"}>
+            <Navbar light className={"pl-0 pr-0 pr-md-4"}>
+                <NavbarToggler onClick={this.toggle} className={"d-md-none"}/>
+                <Collapse isOpen={this.state.isOpen} navbar className={"d-md-block mt-4 mt-md-0"}>
```
MartinEden: As discussed, it's unclear how the hide/show magic works here. Maybe add a comment or wrap it in an element (which itself would probably need a couple of comments)

PR #196, opened 2018-02-22
I1527 - Implement caching for redux stores

georgearrowwood:


app/src/main/shared/actions/authActions.ts
```diff
@@ -98,7 +98,9 @@ export const authActions = {
     logOut() {
         return (dispatch: Dispatch<any>, getState: () => GlobalState) => {
             localStorageHandler.remove("accessToken");
-            (new AuthService(dispatch, getState)).clearShinyCookie();
+            (new AuthService(dispatch, getState))
+                .clearAllCache()
+                .clearShinyCookie();
```
littledice: Being able to chain like this is nice :)

app/src/main/shared/services/AbstractLocalService.ts
```diff
@@ -87,28 +97,54 @@ export abstract class LocalService {
...
+    protected getData(url: string, method: string, params?: any) {
+        if (this.options.cache) {
+            const cacheValue = this.getCache(this.makeUrl(url));
+            if (cacheValue) {
+                // reset options on returning cached data from endpoint
+                this.initOptions();
+                return Promise.resolve(cacheValue);
+            }
+        }
+        return this.doFetch(this.makeUrl(url), this.makeRequestOptions(method, params))
```
littledice: `this.makeUrl(url)` is possibly called twice in this method now, (line 123 and here) so could pull it out into a local variable

app/src/main/shared/services/AbstractLocalService.ts
```diff
@@ -87,28 +97,54 @@ export abstract class LocalService {
...
-        return this.doFetch(this.makeUrl(url), this.makeRequestOptions('POST', params))
+        return this.getData(url, "POST", params);
+    }
+
+    protected getCache(url: string) {
+        return this.cacheEngine.get(["localService", this.constructor.name, this.options.cache, encodeURIComponent(url)].join('.'));
+    }
+
+    protected setCache(url: string, data: any) {
+        this.cacheEngine.set(["localService", this.constructor.name, this.options.cache, encodeURIComponent(url)].join('.'), data);
```
littledice: Lets pull out the logic to generate cache keys in the required format into it's own function, so it can be reused when getting and setting cache values.

app/src/main/shared/services/AbstractLocalService.ts
```diff
@@ -130,33 +166,33 @@ export abstract class LocalService {
...
                     + response.toString(), "error");
         }
     }
 
     notifyOnErrors(error: any) {
+        this.initOptions();
         notificationActions.notify(error);
     }
 
-    errorToNotificationException(error: any): NotificationException {
```
littledice: was this just no longer used?

app/src/test/shared/services/cache/SingletonVariableCache.ts
```diff
@@ -0,0 +1,20 @@
+import { expect } from "chai";
+import { singletonVariableCache } from "../../../../main/shared/services/cache/singletonVariableCache";
+
+describe('Singleton variable cache tests', () => {
+
+    it('sets and gets data', () => {
+        const key = "test.test1.test2";
+        singletonVariableCache.clearAll();
+        expect(singletonVariableCache.get(key)).to.equal(undefined);
```
littledice: I'm finding this test hard to follow - could you split it up into separate tests with one or two `expect` statements each?

app/src/main/shared/services/ModellingGroupsService.ts
```diff
@@ -1,8 +1,9 @@
-import { LocalService } from "../../shared/services/LocalService";
+import { AbstractLocalService } from "./AbstractLocalService";
 
-export class ModellingGroupsService extends LocalService {
+export class ModellingGroupsService extends AbstractLocalService {
 
     getAllGroups() {
-        return this.get("/modelling-groups/");
+        return this.setOptions({cache: 'groups'})
```
littledice: If we ever want to invalidate the cache using this part of the key, might be a good idea to make this option and enum rather than a string, so we can keep track of the keys we're using?

app/src/main/report/services/ReportsService.ts
```diff
@@ -22,3 +25,9 @@ export class ReportsService extends ReportLocalService {
         return this.post(`/reports/${name}/versions/${version}/publish/?value=false`);
     }
 }
+
+export enum ReportsCacheKeysEnum {
+    "reports" = "reports",
+    "versions" = "versions",
+    "versionDetails" = "versionDetails"
+}
```
littledice: Nice, this feels much better and will be handy if we do want to invalidate cache at any of these levels :+1:

PR #197, opened 2018-02-22
I1510 - Sort reports list

georgearrowwood:


app/src/main/report/components/ReportsList/ReportsListSorting.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+interface ReportsListSortingProps {
+    sort: (value: ReportsSortingFields) => void;
+}
+
+export const ReportsListSortingComponent: React.StatelessComponent<ReportsListSortingProps> = (props: ReportsListSortingProps) => (
+    <div>
+        <label>Sorting</label>
+        <select onChange={(e : React.ChangeEvent<HTMLSelectElement>) => props.sort(e.target.value as ReportsSortingFields)} >
+            <option value=""></option>
```
littledice: We don't need an empty default value here; it can default to 'name'

app/src/main/report/components/ReportsList/ReportsListSorting.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { Dispatch, Action } from "redux";
+
+import {reportsActions} from "../../actions/reportsActions";
+import {ReportsSortingFields} from "../../actionTypes/ReportsActionsTypes";
+
+interface ReportsListSortingProps {
+    sort: (value: ReportsSortingFields) => void;
+}
+
+export const ReportsListSortingComponent: React.StatelessComponent<ReportsListSortingProps> = (props: ReportsListSortingProps) => (
```
littledice: Oh nice, didn't know about this `StatelessComponent` type, that's something we can use in a lot of places!

app/src/main/report/components/ReportsList/ReportsListSorting.tsx
```diff
@@ -0,0 +1,31 @@
...
+    <div>
+        <label>Sorting</label>
```
littledice: Notes for styling this:
1. "Sort by:" rather than "Sorting"
2. wrap `label` and `select` in a `form-group`, wrapped in a `form-inline`
3. give the `select` a class of `form-control`

app/src/main/report/reducers/reportsReducer.ts
```diff
@@ -20,6 +21,8 @@ export const reportsReducer = (state = reportsInitialState, action: ReportsActio
     switch (action.type) {
         case ReportTypeKeys.REPORTS_FETCHED:
             return { ...state, reports: action.data };
+        case ReportTypeKeys.REPORTS_SORTED:
+            return { ...state, reports: clone(state.reports).sort((a: Report, b: Report)=> a[action.data] < b[action.data] ? -1 : 1) };
```
littledice: Why do you have to clone `state.reports` here?

app/src/test/report/reducers/ReportsReducersTests.ts
```diff
@@ -28,4 +28,15 @@ describe('Reports reducer tests', () => {
         expect(newState.versionDetails).to.eql(versionMock);
     });
 
+    it('sorts reports list', () => {
+        const reportsMock = [mockReport({name: "b"}), mockReport({name: "c"}), mockReport({name: "a"})];
```
littledice: This tests that they are sorted by name, just need another test to make sure they are sorted by date as expected

app/src/main/report/reducers/reportsReducer.ts
littledice: If I'm reading this logic correctly, sorting by `latest_version` will result in older versions at the top of the list, whereas we want newer versions at the top. Pulling out the sorting logic into its own method would also make this a little easier to read.

app/src/main/report/components/ReportingForgottenPasswordPage.tsx
```diff
@@ -22,7 +22,7 @@ export class ReportingForgottenPasswordPage extends ReportingPageWithHeader<unde
     }
 
     parent(): IPageWithParent {
-        return new MainMenu();
+        return new ReportsListPage();
```
littledice: Good shout - this naming is a lot clearer :+1:

app/src/main/report/actionTypes/ReportsActionsTypes.ts
```diff
@@ -3,6 +3,7 @@ import {Version} from "../../shared/models/reports/Report";
 
 export enum ReportTypeKeys {
     REPORTS_FETCHED = "REPORTS_FETCHED",
+    REPORTS_SORTED = "REPORTS_SORTED",
```
littledice: For this action I'd probably go with `SORT_REPORTS`, because it's dispatched when the user makes a request to sort reports, not after the reports are sorted?

app/src/main/report/components/ReportsList/ReportsList.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export const mapDispatchToProps = (dispatch: Dispatch<ReportAppState>): Partial<ReportsListContainerProps> => {
+    return {
+        getReports: () => dispatch(reportsActions.getReports())
+    }
+};
+
+const enhance = compose(
+    connect(mapStateToProps, mapDispatchToProps),
+    lifecycle(lifecyleProps),
```
littledice: My only qualm here is that now we have 2 styles of component - this style and the component class style. We need to choose whether or not we're going to use `recompose` and make sure we create components consistently going forwards.

I feel like I need to write a component, write some tests etc myself to really get a feel for whether or not it's a better pattern. I propose we give ourselves 1 sprint to try it out, and at the end we decide to either use `recompose` everywhere or not at all. Looping @MartinEden in on this discussion too, when he comes back online!

app/src/main/report/components/ReportsList/ReportsList.ts
```diff
@@ -0,0 +1,50 @@
...
+    ready: boolean;
+}
+
+export const getDisplayedReportsList = (items: Report[], sortBy: ReportsSortingFields) => {
+    if (!items || !items.length) return [];
+    let displayItemsSorted = clone(items).sort(sortReports(sortBy));
+    return displayItemsSorted;
+}
+
+export const sortReports = (sortBy: ReportsSortingFields) => (a: Report, b: Report)=> (a[sortBy] < b[sortBy] ? -1 : 1);
```
littledice: See comment on the tests - this logic will have to vary depending on whether we're sorting by name or by date

app/src/main/report/components/ReportsList/ReportsList.ts
```diff
@@ -0,0 +1,41 @@
...
+import {getDisplayedReportsListSelector} from "./reportsListSelectors";
+
+export interface ReportsListContainerProps extends ReportsListComponentProps {
+    getReports: () => void;
+    ready: boolean;
+}
+
+const lifecyleProps = {
+    componentDidMount() {
+        this.props.getReports();
```
littledice: I'm a little confused here - this logic makes less sense when decoupled from a component. What does `this` refer to, and how do we know that `this.props` exists or has a property called `getReports`? We're just relying on this object being mapped correctly into a component of the right type to hook up the `this` keyword in an appropriate way; is there a way with recompose to enforce this a bit more rigorously?

app/src/test/report/components/ReportsList/ReportsListSelectorsTests.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    it ("selects display list with sorting by latest_version", () => {
+        const reports = [
+            mockReport({name: "b", latest_version: "20170327-002851-dd944766"}),
+            mockReport({name: "c", latest_version: "20170326-002851-dd944766"}),
+            mockReport({name: "a", latest_version: "20170328-002851-dd944766"})
+        ];
+        const displayReports = getDisplayedReportsListSelector(mockReportAppState({reports: {reports, reportsSortBy: ReportsSortingFields.latest_version}}));
+        expect(displayReports[0].name).to.eql("a");
+        expect(displayReports[1].name).to.eql("b");
```
littledice: The `latest_version` ordering coincides with the alphabetical ordering - could we just mix up the letters here so it's clear that it's really being sorted by `latest-version` and not `name`?

app/src/main/report/components/ReportsList/ReportsList.ts
```diff
@@ -0,0 +1,43 @@
...
+import {getDisplayedReportsListSelector} from "./reportsListSelectors";
+
+export interface ReportsListContainerProps extends ReportsListComponentProps {
+    getReports: () => void;
+    ready: boolean;
+}
+
+const lifecyleProps = {
+    onDidMount(props: ReportsListContainerProps) {
+        console.log('pr', props)
```
littledice: This seems better to me now it has the `props` typing, thanks. Is this functionality under test?

app/src/main/report/components/ReportsList/ReportsList.ts
```diff
@@ -0,0 +1,43 @@
...
+export const mapDispatchToProps = (dispatch: Dispatch<ReportAppState>): Partial<ReportsListContainerProps> => {
+    return {
+        getReports: () => dispatch(reportsActions.getReports())
+    }
+};
+
+const enhance = compose(
+    connect(mapStateToProps, mapDispatchToProps),
+    withLifecycle(lifecyleProps),
+    branch((props: ReportsListContainerProps) => !props.ready, renderComponent(LoadingElement))
```
littledice: What's the best way to test `recompose` components? E.g. this branching logic, is this under test?

MartinEden:
I don't have time to review this fully, but it looks good :)

littledice:
There's a build failure that looks unrelated to this changeset, investigating before merging

PR #204, opened 2018-03-02
I1546 - sort list of report readers

littledice:
Also renamed component and added missing tests


app/src/test/report/components/Sidebar/SidebarAdminTests.tsx
```diff
@@ -94,6 +95,26 @@ describe("SidebarAdmin", () => {
         expect(rendered.find(PublishSwitch)).to.have.lengthOf(0);
     });
 
+    it("renders report readers if user is admin", () => {
```
littledice: realised these tests were missing

PR #209, opened 2018-03-13
I1563 - ability to delete local report readers

littledice:
Make local report readers (i.e. with `reports-reader` permission scoped to the current report deletable via the UI.

I also refactored the `SidebarAdmin` component to use `recompose` - still can't decide whether I like it or not!


app/scripts/add-test-accounts-for-integration-tests.sh
```diff
@@ -22,3 +22,4 @@ $here/cli.sh add "Report reviewer" report.reviewer \
 $here/cli.sh addRole report.reviewer user
 $here/cli.sh addRole report.reviewer reports-reviewer
 $here/cli.sh addRole report.reviewer user-manager
+$here/cli.sh addRole report.reviewer reports-reader report:html
```
littledice: Adding a local report reader as well as a global one, for testing

app/src/test/shared/services/LocalServiceTests.ts
```diff
@@ -145,77 +150,98 @@ describe('Local service cache tests', () => {
 
     const sandbox = new Sandbox();
 
+    class TestService extends AbstractLocalService {
```
littledice: Factored out logic shared between the tests to make the individual tests easier to read.

app/src/main/shared/services/AbstractLocalService.ts
```diff
@@ -105,9 +105,13 @@ export abstract class AbstractLocalService {
         return this.getData(this.makeUrl(url), "POST", params);
     }
 
-    protected makeCacheKey(url: string) : string {
-        if (!url || !this.options.cacheKey) return null;
```
littledice: I removed this check as I can't see why it would be needed.

PR #215, opened 2018-03-15
I1604 - render image artefacts inline

littledice:
Included:
* new inline artefact figure component, for rendering images inline
* styled the download sections as per the mockups

Not included:
* make sections collapsible (https://vimc.myjetbrains.com/youtrack/issue/VIMC-1609)


app/package.json
```diff
@@ -113,7 +113,7 @@
     "react-router-redux": "^5.0.0-alpha.9",
     "react-scrollable-anchor": "^0.5.0",
     "react-transition-group": "^1.2.0",
-    "reactstrap": "^4.8.0",
+    "reactstrap": "5.0.0-beta",
```
littledice: Although this is in beta it contains some bug fixes from the previous version (in particular, the `Card` components work as expected now) and is the version referenced in the docs.

app/src/main/report/components/Artefacts/ArtefactRow.tsx
```diff
@@ -1,18 +0,0 @@
-import * as React from "react";
```
littledice: This class is obsolete. It has effectively been replaced by the `ReportDownloadSection` component which is used for all downloads, not just artefacts

PR #217, opened 2018-03-16
i1605 - switched to redux form from alt form for forgot password

georgearrowwood:
switched to redux form from alt form for forgot password


georgearrowwood:
its also on contrib and admin

app/src/main/shared/components/Login/ForgottenPasswordForm.tsx
```diff
@@ -1,25 +1,69 @@
...
+}
 
-    render() {
+export interface ForgotPasswordFormFields{
+    email: string;
+}
+
+export class ForgottenPasswordFormComponent extends React.Component<ForgotPasswordFormProps, undefined> {
+    renderField(data: any) {
+        const { input, label, type, meta: { touched,  error } } = data;
```
littledice: This looks like nice structured data; is there a type for it?

PR #218, opened 2018-04-04
1. I1618 - contrib redux-ation

georgearrowwood:


app/src/main/contrib/actionTypes/CoverageTypes.ts
```diff
@@ -0,0 +1,33 @@
...
+    COVERAGE_ONE_TIME_TOKEN_CLEAR = "COVERAGE_ONE_TIME_TOKEN_CLEAR"
+}
+
+export interface CoverageDataSetsFetched {
+    type: CoverageTypes.COVERAGE_DATA_SETS_FETCHED;
+    data: CoverageSet[];
+}
+
+export interface CoverageSetFormat {
+    type: CoverageTypes.COVERAGE_SET_FORMAT;
```
littledice: A `CoverageSet` is a concept in our data model, so using `Set` as a verb here is a bit confusing! 

In general, if we're prefixing interface names in this way (and I agree it might be confusing to have 2 actions called `SetFormat`, one for demographics and one for coverage) I think the ideal way to do this is with namespaces:
https://www.typescriptlang.org/docs/handbook/namespaces.html

So this would become
```
export namespace Coverage {
    export interface DataSetsFetched {
        type: CoverageTypes.COVERAGE_DATA_SETS_FETCHED;
        data: CoverageSet[];
    }

    export interface SetFormat {
        type: CoverageTypes.COVERAGE_SET_FORMAT;
        data: string;
    }

    export interface OneTimeTokenFetched {
        type: CoverageTypes.COVERAGE_ONE_TIME_TOKEN_FETCHED;
        data: string;
    }

    export interface OneTimeTokenClear {
        type: CoverageTypes.COVERAGE_ONE_TIME_TOKEN_CLEAR;
    }
}

export type CoverageAction =
    | Coverage.DataSetsFetched
    | Coverage.SetFormat
    | Coverage.OneTimeTokenFetched
    | Coverage.OneTimeTokenClear
```

Then in `coverageActioncreators` you just need to import this namespace, and you can reference the actions individually in the same way (`Coverage.SetFormat` etc)

This seems like a sensible pattern for us to adopt in general, although particularly useful in the case of coverage and demographics; what do you think @MartinEden ?

app/src/main/contrib/actionTypes/RunParametersTypes.ts
```diff
@@ -0,0 +1,43 @@
...
+    data: ModelRunParameterSet[];
+}
+
+export interface RunParametersTokenFetched {
+    type: RunParametersTypes.RUN_PARAMETERS_TOKEN_FETCHED;
+    data: tokenData;
+}
+
+export enum RunParametersUploadStatus {
+    off = 'off' ,
```
littledice: Could you explain how these statuses are used? Do we need 3 statuses or can we just have a boolean value that is set to true when the upload is complete?

app/src/main/contrib/actions/coverageActionCreators.ts
```diff
@@ -0,0 +1,59 @@
...
+    getOneTimeToken() {
+        return async (dispatch: Dispatch<any>, getState: any) => {
+            dispatch({
+                type: CoverageTypes.COVERAGE_ONE_TIME_TOKEN_CLEAR,
+            } as CoverageOneTimeTokenClear );
+
+            const touchstone = getState().touchstones.currentTouchstone;
+            const format = getState().coverage.selectedFormat;
+            const group = getState().groups.currentUserGroup;
+            const responsibility = getState().responsibilities.currentResponsibility;
```
littledice: Looks like touchstone id, group id and scenario id are used both here and in the above function, so would be nice to refactor this out into a re-usable function like
```
function getResponsibilityIds(state: ContribAppState){
    const groupId = state.groups.currentUserGroup.id;
    const touchstoneId = state.touchstones.currentTouchstone.id;
    const scenarioId = state.responsibilities.currentResponsibility.scenario.id;
    return {groupId, touchstoneId, scenarioId}
}
```
This can be called like
```
const ids = getResponsibilityIds(getState());
 const token: string = await (new CoverageService(dispatch, getState))
+                .getOneTimeToken(ids.groupId, ids.touchstoneId, ids.scenarioId, format);

```

I think these 3 ids are also used in other action creators (e.g. I see them also in `runParamerersActionCreators`) so this function could live in some helper class accessible to all of them.

MartinEden:
I've reviewed all the action creators, and they look good. :+1:

app/src/main/contrib/components/Responsibilities/BurdenEstimates/CreateBurdenEstimateSetForm.tsx
```diff
@@ -74,35 +82,49 @@ export class CreateBurdenEstimateSetForm extends React.Component<BurdenEstimateP
...
+                               {/*onChange={this.onDetailsChange.bind(this)}/>*/}
+                        {/*<small className="form-text text-muted">*/}
+                            {/*For example, if averaged, what kind of averaging function was used.*/}
+                        {/*</small>*/}
+                        {/*<div className="invalid-feedback">*/}
+                            {/*Please give us some details, a short note is fine*/}
+                        {/*</div>*/}
+                    {/*</div>*/}
+                {/*</div>*/}
+            {/*</Form>*/}
```
MartinEden: This is unfinished, right?

app/src/main/contrib/components/Responsibilities/Coverage/DownloadCoveragePage.tsx
```diff
@@ -1,52 +1,64 @@
...
+        </PageArticle>
     }
+}
 
-    parent(): IPageWithParent {
-        return new ResponsibilityOverviewPage();
+export const mapStateToProps = (state: ContribAppState): Partial<DownloadCoveragePageProps> => {
+    console.log('pp', state)
+    return {
+        responsibility: state.responsibilities.currentResponsibility,
```
littledice: Where is this prop needed/used?

app/src/main/contrib/reducers/responsibilitiesReducer.ts
```diff
@@ -0,0 +1,24 @@
+import {ResponsibilitiesAction, ResponsibilitiesTypes} from "../actionTypes/ResponsibilitiesTypes";
+import {ExtendedResponsibility, IExtendedResponsibilitySet} from "../models/ResponsibilitySet";
+
+export interface ResponsibilitiesState {
+    set: IExtendedResponsibilitySet;
```
littledice: can we rename this to `responsibilitySet`? `set` is too ambiguous

app/src/main/contrib/reducers/responsibilitiesReducer.ts
```diff
@@ -0,0 +1,24 @@
...
+export const initialState: ResponsibilitiesState = {
+    set: null,
+    currentResponsibility: null,
+};
+
+export const responsibilitiesReducer = (state = initialState, action: ResponsibilitiesAction) => {
+    switch (action.type) {
+        case ResponsibilitiesTypes.SET_RESPONSIBILITIES:
+            return {...state, set: action.data };
+        case ResponsibilitiesTypes.SET_CURRENT_RESPONSIBILITY_SET:
```
littledice: The action type here is called `SET_CURRENT_RESPONSIBILITY_SET` but actually the logic is setting the current responsibility, not the current responsibility set! I think the logic is right, and the action type should be `SET_CURRENT_RESPONSIBILITY`

app/src/main/contrib/components/Responsibilities/Overview/List/ResponsibilityScenario.tsx
```diff
@@ -1,25 +1,30 @@
...
 import {CurrentEstimateSetSummary} from "./CurrentEstimateSetSummary";
+import {ContribAppState} from "../../../../reducers/contribAppReducers";
 
-interface Props {
+export interface ResponsibilityScenarioProps {
     responsibility: Responsibility;
     modellingGroup: ModellingGroup;
     touchstone: Touchstone;
     responsibilitySetStatus: ResponsibilitySetStatus;
+    diseases?: Disease[];
```
littledice: I would prefer if we used the `PublicProps` approach we've used elsewhere. So 
```
interface ResponsibilityScenarioPublicProps {
     responsibility: Responsibility;
     modellingGroup: ModellingGroup;
     touchstone: Touchstone;
     responsibilitySetStatus: ResponsibilitySetStatus;	     
     responsibilitySetStatus: ResponsibilitySetStatus;
}

interface ResponsibilityScenarioProps  extends ResponsibilityScenarioPublicProps {
     diseases: Disease[];
}
```

Then `mapStateToProps` takes an argument of type `ResponsibilityScenarioPublicProps` and returns  `ResponsibilityScenarioProps`

littledice: Similarly in the `TemplateLinks` component below

PR #221, opened 2018-04-16
i1636 - turn rfp touchstone ids into array

littledice:
and add new rfp touchstone id


littledice:
Yeah you're right, they should all have "rfp-" in them

PR #222, opened 2018-04-16
2. I1618 unit tests

georgearrowwood:


app/src/test/mocks/mocks.ts
```diff
@@ -41,35 +41,35 @@ export function mockHistory(params?: any) {
...
-        MainStore: {
-            diseases: makeLoadable<models.Disease>(config.diseases),
-            modellingGroups: makeLoadable<models.ModellingGroup>(config.groups),
-        },
-        ResponsibilityStore: {
-            touchstones: config.touchstones,
-        }
-    }));
-}
+// export function setupMainStore(config: {
```
littledice: Can we delete all this commented code?

app/src/test/contrib/components/Responsibilities/Uploads/UploadEstimatesContentTests.tsx
```diff
@@ -1,136 +0,0 @@
...
-import {Sandbox} from "../../../../Sandbox";
-import {
-    UploadBurdenEstimatesContentComponent,
-    UploadBurdenEstimatesContentComponentProps
-} from "../../../../../main/contrib/components/Responsibilities/BurdenEstimates/UploadBurdenEstimatesContent";
-import {CurrentEstimateSetSummary} from "../../../../../main/contrib/components/Responsibilities/Overview/List/CurrentEstimateSetSummary";
-import {UploadBurdenEstimatesForm} from "../../../../../main/contrib/components/Responsibilities/BurdenEstimates/UploadBurdenEstimatesForm";
-
-describe("UploadEstimatesContentComponent", () => {
-    const sandbox = new Sandbox();
```
littledice: Shouldn't all of these tests, or some equivalents, still exist?

app/src/test/contrib/components/ContribNoRouteFoundPageTests.tsx
```diff
@@ -0,0 +1,36 @@
...
+    });
+
+    it("renders component on component level", () => {
+        let store = createMockStore();
+        const createBreadcrumbsStub = sandbox.setStubReduxAction(breadcrumbsActionCreators, "createBreadcrumbs");
+        const rendered = shallow(<ContribNoRouteFoundPage/>, {context: {store}}).dive().dive();
+        const pageArticle = rendered.find('PageArticle');
+        expect(createBreadcrumbsStub.called).is.equal(true);
+        expect(pageArticle.props().title).is.equal(NoRouteFound.title());
+        console.log(pageArticle.debug());
```
littledice: Huh, this `.debug()` function is useful - I didn't know it existed! But can probably remove this console log now :)

app/src/main/shared/components/OneTimeButton/OneTimeButton.tsx
```diff
@@ -31,7 +32,11 @@ export class OneTimeButton extends React.Component<OneTimeButtonProps, any> {
     }
 
     refreshToken() {
-        setTimeout(this.props.refreshToken);
+        if (typeof this.props.refreshToken === "function") {
+            setTimeout(() => {
+                this.props.refreshToken();
+            });
```
MartinEden: Why not just:
```
if (typeof this.props.refreshToken === "function") {
    setTimeout(this.props.refreshToken);
}
```

app/src/test/contrib/actions/CoverageActionsTests.ts
```diff
@@ -0,0 +1,80 @@
+import { expect } from "chai";
+
+import { Sandbox } from "../../Sandbox";
+import { coverageActionCreators } from "../../../main/contrib/actions/coverageActionCreators";
+import { CoverageService } from "../../../main/contrib/services/CoverageService";
+import {Coverage, CoverageTypes} from "../../../main/contrib/actionTypes/CoverageTypes";
+import {createMockStore} from "../../mocks/mockStore";
+import {mockCoverageSet, mockDemographicDataset, mockScenario, mockTouchstone} from "../../mocks/mockModels";
+import {ScenarioAndCoverageSets} from "../../../main/shared/models/Generated";
+import {statePropsMapHelper} from "../../../main/contrib/helpers/statePropsMapHelper";
```
MartinEden: I didn't catch this on the first PR, but I find this name rather unclear. Could you perhaps rename it to `mapStateToPropsHelper`?

app/src/test/contrib/actions/CoverageActionsTests.ts
```diff
@@ -0,0 +1,80 @@
...
+    };
+
+    afterEach(() => {
+        sandbox.restore();
+    });
+
+    it("data sets fetched", (done) => {
+        const store = createMockStore({});
+        sandbox.setStubFunc(statePropsMapHelper, "getResponsibilityIds", ()=>{
+            return Promise.resolve({groupId: "g1", touchstoneId: "t1", scenarioId: "s1"});
```
MartinEden: Why does this need to be a promise? The original function is synchronous.

app/src/test/contrib/actions/CoverageActionsTests.ts
```diff
@@ -0,0 +1,80 @@
...
+        sandbox.setStubFunc(CoverageService.prototype, "getDataSets", ()=>{
+            return Promise.resolve(testScenarioAndCoverageSet);
+        });
+
+        store.dispatch(coverageActionCreators.getDataSets())
+        setTimeout(() => {
+            const actions = store.getActions();
+            const expectedPayload = { type: CoverageTypes.COVERAGE_DATA_SETS_FETCHED, data: [testCoverageSet] };
+            expect(actions).to.eql([expectedPayload]);
+            done();
```
MartinEden: So previously we had the `checkAsync` helper for this kind of check. I strongly recommend we continue to use this throughout, as it handles the case where an exception is thrown by calling `done(exception)`. Otherwise, if this assertion fails, the test will hang for several seconds and then timeout without us getting proper feedback. Unless mocha has upgraded this behaviour?

app/src/test/contrib/actions/CoverageActionsTests.ts
```diff
@@ -0,0 +1,80 @@
...
+    it("sets format", (done) => {
+        const store = createMockStore({});
+        store.dispatch(coverageActionCreators.setFormat(Coverage.SelectedFormat.long));
+        setTimeout(() => {
+            const actions = store.getActions();
+            const expectedPayload = { type: CoverageTypes.COVERAGE_SET_FORMAT, data: "long" };
+            expect(actions).to.eql([expectedPayload]);
+            done();
+        });
+    });
```
MartinEden: Nice, very easy to read.

app/src/test/contrib/actions/DemographicActionsTests.ts
```diff
@@ -0,0 +1,92 @@
...
+        });
+    });
+
+    it("set gender", (done) => {
+        const store = createMockStore({});
+        store.dispatch(demographicActionCreators.setGender('male'))
+        setTimeout(() => {
+            const actions = store.getActions()
+            const expectedPayload = { type: DemographicTypes.DEMOGRAPHIC_SET_GENDER, data: 'male' }
+            expect(actions).to.eql([expectedPayload])
```
MartinEden: There's some boiler plate repeated through these tests. Frequently: 1. get the actions 2. state the expectation 3. check the actions equals singleton array of expected payload. It would be better to just have (as we had previously) a helper along the lines of 

`expectOneAction({ type: DemographicTypes.DEMOGRAPHIC_SET_GENDER, data: 'male' })`

This will remove a fair few identical lines.

app/src/test/contrib/actions/pages/DownloadCoveragePageActionsTests.ts
```diff
@@ -0,0 +1,111 @@
...
+                { type: ModellingGroupTypes.SET_CURRENT_USER_GROUP, data: testGroup },
+                { type: TouchstoneTypes.TOUCHSTONES_FETCHED, data: [testTouchstone]},
+                { type: DiseasesTypes.DISEASES_FETCHED, data: [testDisease]},
+                { type: TouchstoneTypes.SET_CURRENT_TOUCHSTONE, data: testTouchstone},
+                { type: ResponsibilitiesTypes.SET_RESPONSIBILITIES, data: testExtResponsibilitySet},
+                { type: ResponsibilitiesTypes.SET_CURRENT_RESPONSIBILITY, data: testResponsibility},
+                { type: CoverageTypes.COVERAGE_ONE_TIME_TOKEN_CLEAR},
+                { type: CoverageTypes.COVERAGE_ONE_TIME_TOKEN_FETCHED, data: "test-token"},
+                { type: CoverageTypes.COVERAGE_DATA_SETS_FETCHED, data: [testCoverageSet]},
+                { type: BreadcrumbsTypes.BREADCRUMBS_RECEIVED, data: testBreadcrumbs },
```
MartinEden: Wow

app/src/test/contrib/components/ChooseAction/ChooseActionPageTests.tsx
```diff
@@ -1,63 +1,41 @@
...
-                mockModellingGroup({ id: "group-2" }),
-            ])
-        });
-        bootstrapStore(responsibilityStore, {
-            currentModellingGroup: mockModellingGroup({id: "group-1"}),
-        });
-        mockFetcherForMultipleResponses([
-            mockTouchstonesEndpoint([mockTouchstone(), mockTouchstone()], "group-1")
-        ]);
+    it("renders component component level", () => {
```
MartinEden: What does this test name mean?

app/src/test/contrib/actions/DemographicActionsTests.ts
georgearrowwood: testing helpers in not a good idea in general, hides logic, makes things obscure

app/src/test/contrib/actions/CoverageActionsTests.ts
georgearrowwood: not , necessary, doesn't hang now

app/src/test/contrib/actions/CoverageActionsTests.ts
georgearrowwood: this was copy paste from somewhere, will need to be fixed, yes

app/src/test/contrib/actions/CoverageActionsTests.ts
georgearrowwood: ok

app/src/main/shared/components/OneTimeButton/OneTimeButton.tsx
georgearrowwood: left from testing something, yeah, will put it back, as it was before

app/src/test/contrib/components/ContribNoRouteFoundPageTests.tsx
georgearrowwood: oh, yeah, will removed that console log
i noticed that component NoRouteFound is strangely rendered
need to redo this too, but it used in too many places
so next time

app/src/test/contrib/actions/CoverageActionsTests.ts
MartinEden: Great :)

app/src/test/contrib/actions/CoverageActionsTests.ts
MartinEden: This still needs to be fixed, I think?

app/src/test/contrib/components/ChooseAction/ChooseActionPageTests.tsx
```diff
@@ -25,7 +25,7 @@ describe("Choose Action Page Component", () => {
         expect(typeof rendered.props().onLoad).is.equal('function');
     });
 
-    it("renders component component level", () => {
+    it("renders component on component level, renders title and sub component", () => {
```
MartinEden: I'm still not sure I get this. Is it meant to just say "on component level, renders title and sub component"?

MartinEden:
Two tiny fixes left! :)

app/src/test/contrib/components/ChooseAction/ChooseActionPageTests.tsx
georgearrowwood: Now we have this redux/recompose/hoc aproach, we have several render levels on almost each component, especially pages and content components. This render path can be for example, and most common: Connect->branch->Component itself.  Shallow render of enzyme does exactly one level, then if you do .dive() after it, it goes deeper to a next level. In this case component level is next after connect. In many tests its not necessarily first level of component, because sometimes we need to go deeper.

PR #223, opened 2018-04-18
I1668

littledice:
Add confidentiality agreement checkbox.

First merge https://github.com/vimc/montagu-db/pull/70
and https://github.com/vimc/montagu-api/pull/260 and return api/db references to master.


app/src/main/contrib/components/Responsibilities/Overview/ConfidentialityAgreement.tsx
```diff
@@ -0,0 +1,75 @@
...
+    signed: boolean;
+    getConfidentiality: () => void;
+}
+
+const fullConfidentialityAgreement = require('./rfp-applicants-confidentiality.pdf');
+
+export const ConfidentialityAgreementComponent: React.SFC<ConfidentialityAgreementProps>
+    = (props: ConfidentialityAgreementProps) => {
+
+    if (settings.isApplicantTouchstone(props.touchstoneId)) {
```
MartinEden: I think this check is redundant because of composing with `branch` below.

app/src/main/contrib/components/Responsibilities/Overview/ConfidentialityAgreement.tsx
```diff
@@ -0,0 +1,75 @@
...
+                    }
+                </div>
+            </div>
+        </div>
+    }
+};
+
+const lifecyleProps: Partial<LifecycleMethods<ConfidentialityAgreementProps>> = {
+    onDidMount(props: ConfidentialityAgreementProps) {
+        props.getConfidentiality();
```
MartinEden: Presumably there is a delay here while this is fetched. Until then, regardless of whether or not the user has signed before won't they be presented with this checkbox?

MartinEden:
Hi @littledice, I've just showed this to Kim, and she had some feedback. First of all she thinks it looks great and says thank you. :)

A few things it would be good to change:

1. It's a little unclear that the box gets ticked because it vanishes so quickly, but Kim does think that a checkbox is a good way to signal agreement. So maybe a checkbox and then a button that says "Continue" that you can't click until you've checked the box?
1. It would be good to hide the agreement section once the user has agreed, instead of displaying it all the time.
1. Finally, and this one could probably slip beyond the Tuesday deadline if necessary, since I told Kim it would be more technically complicated: Currently you can subvert the agreement just be going to any direct link that is part of the touchstone. e.g. if someone in their team who has agreed shares a link to a particular coverage page. We should probably add a composable redirect thingybob to all the pages that are part of a touchstone. That might have to be a new ticket.

app/src/main/contrib/actions/userActionCreators.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    signConfidentialityAgreement() {
+
+        return async (dispatch: Dispatch<any>, getState: () => GlobalState) => {
+            const result: any = await (new UserService(dispatch, getState)).signConfidentiality();
+
+            if (result == "OK") {
+                dispatch({
+                    type: UserActionType.CONFIDENTIALITY_SIGNED,
+                    data: true
```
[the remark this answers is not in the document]
littledice: No you're right, I don't see why all actions should have to have the `data` property if its not used.

app/src/main/contrib/components/Responsibilities/Overview/ConfidentialityAgreement.tsx
littledice: Yes, good catch, thanks

app/src/main/contrib/components/Responsibilities/Overview/ConfidentialityAgreement.tsx
littledice: Yes, in practice this is a very short delay, but you're right. Have changed default state of the `signed` variable to null, so this only gets rendered once the value has been fetched.

littledice:
Ok, have turned this into a HOC which wraps the coverage download page as well as the responsibilities overview page, and I think it should now have the look and feel Kim is after.

app/src/main/contrib/components/Responsibilities/Overview/ConfidentialityAgreement.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+const lifecyleProps: Partial<LifecycleMethods<ConfidentialityProps>> = {
+    onDidMount(props: ConfidentialityProps) {
+        props.getConfidentiality();
+    }
+};
+
+const renderConfidentialityAgreementOrLoadingElement =
+    branch((props: ConfidentialityPropsFromState) => props.signed == false,
+        renderComponent(ConfidentialityAgreementComponent),
```
georgearrowwood: would be nice to have only one branch result here in case if it not stisfy conditions, loading component

georgearrowwood: i noticed branch has one problem, if component has itself triggering of data load, it will not work(such as triggering load on componend load), it only works if you receive props from outside, from parent, or from other component, otherwise it will replace itself with loading component and stop functioning, workaround for this would be using containers pattern: 
https://medium.com/@learnreact/container-components-c0e67432e005

littledice: Re first comment, unfortunately this can't be simplified in that way because it's a ternary condition: 
1. if the touchstone is non rfp, or the agreement has been signed, render the wrapped component.
Otherwise,
2. if `signed` is null render loading component
3. if `signed` is false, render the agreement form

But maybe there's a way I can make this a bit clearer; maybe using recompose here obfuscates things?

littledice: I'm afraid I don't understand the second comment. Do you mean that we shouldn't make the call to `getConfidentiality` from this component? The lifecycle method is hooked up before the branching logic, so I don't think this is actually a problem unless I'm missing something? Could you provide steps to replicate the issue you're describing? We could of course move the call to `getConfidentiality` to the pages that use this component. it just means remembering to always make that call before using this HOC.

littledice: Re the second, I think you're referring to the all to `props.getConfidentiality` in the `ConfidentialityAgreement.componentDidMount` method - that's actually erroneous, so thanks for spotting! The call to `gteConfidentiality` is already mapped to `componentDidMount` in the HOC using recompose (which happens before the branching logic so won't be an issue). I've deleted the obsolete call.

app/src/main/contrib/components/Responsibilities/Overview/ConfidentialityAgreement.tsx
```diff
@@ -0,0 +1,108 @@
...
+    onChange(e: ChangeEvent<HTMLInputElement>) {
+
+        this.setState({
+            checked: e.target.checked
+        })
+    }
+
+    render() {
+
+        if (this.props.signed == null) {
```
littledice: Have moved half of the branching logic out of recompose and into the component's render method. Hopefully this is easier to follow now!

app/src/main/contrib/components/Responsibilities/Overview/ConfidentialityAgreement.tsx
littledice: Okay, moved the second half of the branching logic out of recompose and into the `ConfidentialityAgreement` component, which hopefully makes this easier to follow!

app/src/main/contrib/components/Responsibilities/Overview/ConfidentialityAgreement.tsx
```diff
@@ -31,22 +31,29 @@ export class ConfidentialityAgreementComponent extends React.Component<Confident
...
-                           onChange={this.onChange}/>
-                    {this.state.checked &&
-                    <button className="btn-success" onClick={this.props.signAgreement}>Submit</button>
-                    }
+
+        if (this.props.signed == null) {
+            return <LoadingElement />
+        }
+        else {
+            return <div className={"row"}>
```
MartinEden: Maybe add a comment here explaining that `signed` must be `false` this point, because if it was `true` the branch would have hidden this component? And then I'm happy for this to merge (assuming TeamCity passes)

PR #226, opened 2018-04-30
put lifecycle method back into confidentiality agreement

littledice:


app/src/main/contrib/components/Responsibilities/Overview/ResponsibilityOverviewPage.tsx
```diff
@@ -38,9 +38,6 @@ export class ResponsibilityOverviewPageComponent extends React.Component<Respons
     render(): JSX.Element {
         if (this.props.touchstone) {
             return <PageArticle title={`Responsibilities in ${this.props.touchstone.description }`}>
-                <ResponsibilityOverviewDescription
-                    currentTouchstoneId={this.props.touchstone.id}
```
littledice: Moved this to inside the `ResponsibilityOverviewContent` component, which gets wrapped with the confidentiality agreement

PR #228, opened 2018-04-30
I1634 - enable stochastic estimates for new rfp touchstone

littledice:


app/src/main/contrib/components/Responsibilities/Overview/ResponsibilityOverviewDescription.tsx
```diff
@@ -8,30 +8,73 @@ interface ResponsibilityOverviewDescriptionProps {
...
-                <a href={`mailto:${settings.vimcEmail}`}>
-                    {settings.vimcEmail}
-                </a>
-            </div>;
+                    If you have any questions or anything is not as you expected, please email&nbsp;
+                    <a href={`mailto:${settings.vimcEmail}`}>
+                        {settings.vimcEmail}
+                    </a>
+                </div>;
+            }
```
MartinEden: It might be nice to break these two different bits of guidance into separate components (probably overkill) or just separate functions. So the logic can look more like "if special touchstone { renderSpecialTouchstoneGuidance() } else { renderNormalRFPGuidance() }, which will be easier to read I think.

app/src/@types/Settings.d.ts
```diff
@@ -6,6 +6,7 @@ interface Settings {
     vimcEmail: string;
     slackUrl: string;
     isApplicantTouchstone: (id: string) => boolean;
+    is201804ApplicantTouchstone: (id: string) => boolean;
```
MartinEden: I know this is just temporary, but maybe it could be `isStochasticApplicantTouchstone` and then later we could make the logic more general?

app/src/main/contrib/components/Responsibilities/Overview/List/TemplateLinks.tsx
```diff
@@ -130,8 +130,18 @@ export class TemplateLink extends React.Component<TemplateLinkProps, undefined>
 
         if (settings.isApplicantTouchstone(this.props.touchstoneId)) {
             return <div>
+                <div>
                     <a key={`central_burden_template_${disease.id}-generic.csv`}
-                       href={`${templatePath}central_burden_template_${disease.id}-generic.csv`}>{disease.name}</a>
+                       href={`${templatePath}central_burden_template_${disease.id}-generic.csv`}>{disease.name}
+                        { settings.is201804ApplicantTouchstone(this.props.touchstoneId) && " - central"}</a>
+                </div>
+                { settings.is201804ApplicantTouchstone(this.props.touchstoneId) &&
```
MartinEden: I don't mind the `{ predicate && renderLogic }` approach for one liners, as above, but for this maybe put the render logic into a separate function?

PR #229, opened 2018-05-03
I1729 Admin groups Redux

georgearrowwood:
Main purpose of this pull request is mainly switching to redux for modelling groups functionality in admin part of webapps, this incudes: modelling groups list, modelling groups details, modelling groups  members. Did also some refactoring on components, to make them look same style and have same style names. All covered with tests.


MartinEden:
I was somehow expecting this one to be smaller.

app/src/main/admin/actionTypes/ModellingGroupsTypes.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export enum ModellingGroupTypes {
+    ADMIN_GROUPS_FETCHED = "ADMIN_GROUPS_FETCHED",
+    ADMIN_GROUP_DETAILS_FETCHED = "ADMIN_GROUP_DETAILS_FETCHED",
+    SET_CURRENT_ADMIN_GROUP = "SET_CURRENT_ADMIN_GROUP",
+    SET_CURRENT_ADMIN_GROUP_MEMBERS = "SET_CURRENT_ADMIN_GROUP_MEMBERS"
+}
+
+export interface AdminGroupsFetched {
+    type: ModellingGroupTypes.ADMIN_GROUPS_FETCHED;
```
littledice: I'm sure you've just ported his terminology over from our code, but I have to say I'm not sure what 'admin group' means; any ideas @MartinEden ? Is it supposed to distinguish all groups from the current logged in user's groups? or is it informal namespacing again, to distinguish `admin` from `contrib`? I wonder if we can improve on these names...

app/src/main/admin/actions/modellingGroupsActionCreators.ts
```diff
@@ -0,0 +1,100 @@
...
+        return (dispatch: Dispatch<ContribAppState>, getState: () => ContribAppState) => {
+            (new ModellingGroupsService(dispatch, getState)).clearCacheForGroupDetails(groupId);
+        }
+    },
+
+    setCurrentGroupMembers() {
+        return (dispatch: Dispatch<AdminAppState>, getState: () => AdminAppState) => {
+            const currentGroupDetails = getState().groups.currentGroupDetails;
+            const allUsers = getState().users.users;
+            if (!currentGroupDetails || !allUsers.length) {
```
littledice: So if `currentGroupDetails` exists or `allUsers` exists, this won't dispatch; why do we check this?

app/src/main/admin/components/MainMenu/MainMenuNew.tsx
```diff
@@ -0,0 +1,39 @@
+import * as React from "react";
```
littledice: What's stopping us replacing the old `MainMenu` component with this new one?

app/src/main/admin/components/ModellingGroups/List/ModellingGroupsListContent.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { ModellingGroupListItem } from "./ModellingGroupListItem";
+
+import {AdminAppState} from "../../../reducers/adminAppReducers";
+
+interface ModellingGroupsProps {
+    groups: ModellingGroup[]
+}
+
+export const ModellingGroupsListContentComponent: React.SFC<ModellingGroupsProps> = (props: ModellingGroupsProps) => {
+    if (Array.isArray(props.groups) && props.groups.length) {
```
littledice: Don't these two conjuncts check for the same thing?

app/src/main/admin/components/MainMenu/MainMenuNew.tsx
georgearrowwood: This file will have to be renamed to MainMenu.tsx and the old one deleted in next PR, this is temporary file to be able to generate breadcrumbs for redux pages.

app/src/main/admin/components/ModellingGroups/SingleGroup/Members/ModellingGroupMembersAdd.tsx
```diff
@@ -89,5 +75,24 @@ export class AddMember extends React.Component<Props, AddMemberState> {
...
+    return {
+        members: props.members,
+        users: state.users.users,
+        groupId: props.groupId
+    }
+};
+
+export const mapDispatchToProps = (dispatch: Dispatch<AdminAppState>): Partial<ModellingGroupMembersAddProps> => {
+    return {
+        addUserToGroup: (groupId: string, username: string) => dispatch(modellingGroupsActionCreators.addUserToGroup(groupId, username))
```
littledice: This is a big improvement :) looks like this component wasn't even under integration test before which was entirely my fault - this was one of the first react component I wrote!

app/src/main/admin/components/MainMenu/MainMenuNew.tsx
georgearrowwood: it will not work with users part, as breadcrumbs are using alt there and will conflict with parent page

app/src/main/admin/actions/modellingGroupsActionCreators.ts
georgearrowwood: it maps users to group details members usernames list, if something doesn't exists, it cannot be mapped, so dispatch empty

georgearrowwood: before that you had this mapping in many places in components, i moved that mapping logic out to actions and reducers, so that we can map it once and use where we need

app/src/main/admin/components/ModellingGroups/List/ModellingGroupsListContent.tsx
georgearrowwood: These are different. first one checks data type, second one the length of it.
[] - is array , but length is 0

littledice: But this just checks whether the length exists, not whether it's equal to 0. Was the latter the intention?

georgearrowwood: no, it's not correct, it not checks if length exists if checks value, so if it is 0, condition will not pass

app/src/main/admin/actions/modellingGroupsActionCreators.ts
georgearrowwood: don't quite understand what you mean. Local variable? Where?

app/src/main/admin/actionTypes/ModellingGroupsTypes.ts
MartinEden: The latter - it's to distinguish the different group fetch capabilities between admin and contrib. May well be redundant now.

app/src/main/admin/components/ModellingGroups/SingleGroup/Members/ModellingGroupMembersList.tsx
```diff
@@ -0,0 +1,44 @@
...
+                groupId={props.groupId}
+                showDelete={props.isAdmin}
+            />)}
+        </div>;
+    }
+};
+
+// TODO: move that logic to reselect if logic becomes more complex
+const mapSortUsers = (users: User[]) => {
+    if (!Array.isArray(users) || !users.length) return [];
```
MartinEden: I'm seeing this pattern repeated. Maybe create a global function along the lines of `isNonEmptyArray`?

app/src/main/admin/reducers/modellingGroupsReducer.ts
```diff
@@ -0,0 +1,31 @@
...
+    groups: [],
+    currentGroup: null,
+    currentGroupDetails: null,
+    currentGroupMembers: []
+};
+
+export const modellingGroupsReducer = (state = modellingGroupInitialState, action: ModellingGroupsAction) => {
+    switch (action.type) {
+        case ModellingGroupTypes.ADMIN_GROUPS_FETCHED:
+            return {...state, groups: action.data ? action.data : [] };
```
MartinEden: Why do we need this guard to make sure it's an array?

georgearrowwood: just to keep up with initial data format of reducers, for example we have some list , and by some reason instead of array some null or undefined was passed to it, that will help keeping , it as empty array. It's not "must" logic, it can be different here, that is just something i have chosen to minimise possible errors.

app/src/main/admin/components/ModellingGroups/SingleGroup/Members/ModellingGroupMembersList.tsx
georgearrowwood: can be done if you wish to

app/src/main/admin/reducers/modellingGroupsReducer.ts
georgearrowwood: but here i was only checking if not empty, maybe i should check if array too

app/src/main/admin/actionTypes/ModellingGroupsTypes.ts
littledice: Ok, in that case I'd suggest we drop the `Admin` prefix and see how we get on? We can revisit using ts namespaces if we run into confusion down the line?

app/src/main/admin/actions/modellingGroupsActionCreators.ts
littledice: Sorry, should have been more clear:

```
            const currentGroupDetails = getState().groups.currentGroupDetails;
            const allUsers = getState().users.users;
            const currentGroupMembers = [];

            if (currentGroupDetails && allUsers.length) {
                currentGroupMembers = compact(currentGroupDetails.members
                .map(memberUsername => allUsers.find(user => user.username === memberUsername))
            }

             dispatch({
                type: ModellingGroupTypes.SET_CURRENT_ADMIN_GROUP_MEMBERS,
                data: currentGroupMembers
           } as SetCurrentAdminGroupMembers);
```

app/src/main/admin/components/ModellingGroups/List/ModellingGroupsListContent.tsx
littledice: Oh, javascript falsey magic :laughing:

app/src/main/admin/components/ModellingGroups/SingleGroup/Members/ModellingGroupMembersList.tsx
MartinEden: Yes please.

app/src/main/admin/actionTypes/ModellingGroupsTypes.ts
georgearrowwood: You are both absolutely right, it is kind of redundant and we can drop that Admin | Contrib namespacing in action naming. Were just initial naming, first what came to mind. Actions from different webapps parts should never intercross, as we only initialise one store per app part and trigger actions from only same correspondent app part. Only thing I am not sure now is how TypeScript will behave, but should be fine, will see.

app/src/main/admin/actions/modellingGroupsActionCreators.ts
georgearrowwood: Yes it is a good idea, to chain actions, because it will never will be called independently anyway. Always set members after details fetch. I actually has this idea in mind from the start, then just decided by some reason to do it independently.

PR #230, opened 2018-05-04
show params section if touchstone is non rfp, or is stochastic

littledice:
We should show the stochastic parameters section if the touchstone Id is not an rfp touchstone, OR if it is a stochastic touchstone.

### DIFF
--- a/app/src/main/contrib/components/Responsibilities/Overview/ResponsibilityOverviewContent.tsx
+++ b/app/src/main/contrib/components/Responsibilities/Overview/ResponsibilityOverviewContent.tsx
@@ -23,21 +23,25 @@ export interface ResponsibilityOverviewContentProps {
     touchstoneId: string;
 }
 
+function paramsSection(props: ResponsibilityOverviewContentProps) {
+    if (settings.isStochasticTouchstone(props.touchstoneId) || !settings.isApplicantTouchstone(props.touchstoneId)) {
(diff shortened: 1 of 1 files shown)


app/src/main/contrib/components/Responsibilities/Overview/ResponsibilityOverviewContent.tsx
```diff
@@ -26,35 +26,38 @@ export interface ResponsibilityOverviewContentProps {
...
-        />
+        const paramsSection =
+            settings.isStochasticTouchstone(props.touchstoneId) || !settings.isApplicantTouchstone(props.touchstoneId) ?
+                <div id="params-section">
+                    <div className="largeSectionTitle">Parameters</div>
+                    <div><a key={"params"}
+                            href={stochasticParams}>Download stochastic parameters template</a>
+                    </div>
+                    <ButtonLink href={parametersUrl}>Upload parameters</ButtonLink>
+                </div> : null;
```
MartinEden: Could you turn this into a function in the form:

```
function paramsSection() {
    if (predicate) {
        return <tags>...</tags>
    } else {
        return null;
}
```

I think using the `predicate ? <tags></tags> : null` form is much harder to read when either the predicate or the tags are more than a few words long.

MartinEden:
Looks good, just waiting for TeamCity to pass

PR #231, opened 2018-05-15
I568 - more metadata in report list

littledice:
Should first merge https://github.com/vimc/montagu-reporting-api/pull/34 and return the reporting API to master.

Replaced all custom sorting/filtering logic with this data table component: https://react-table.js.org/#/story/readme

The typings are a little ropey, had to add a few of my own, but the component itself seems good, customisable etc. 

This is a departure from our original designs and renders the search widget obsolete. But with the amount of metadata that users want to search and sort by, this feels like a sensible approach.


littledice:
This PR is way too big. Closing and replacing with smaller ones.

PR #232, opened 2018-05-15
replace report list date filter with a generic date range picker

littledice:
Decouple this component from redux/reports, making it more flexible.


app/src/main/report/components/ReportsList/Filter/ReportsListFilterDate.tsx
```diff
@@ -1,53 +0,0 @@
-import * as React from "react";
-
-import {ReportsFilterFields} from "../../../actionTypes/ReportsActionsTypes";
-import {DatePicker} from "../../../../shared/components/DatePicker/DatePicker";
-import {CalendarIcon} from "../Calendar";
-
-interface ReportsListFilterProps {
-    filterData: ReportsFilterFields;
```
littledice: The key difference between this and the new component is this `filterData` property, which tied it quite closely to the reports redux implementation, has been replaced with a `value` property that just has a start and end date.

app/src/main/report/components/ReportsList/Filter/ReportsListFilterDate.tsx
```diff
@@ -1,53 +0,0 @@
...
-    timeFromSelected: (date: Date) => void;
-    timeUntilSelected: (date: Date) => void;
-}
-
-const fromMonth = new Date("2017-03-01T00:00:00");
-const toMonth = new Date;
```
littledice: Also pulled these values out of this component, instead passing them in as props.

app/src/main/report/components/ReportsList/Filter/ReportsListFilterDate.tsx
littledice: See new props interface: https://github.com/vimc/montagu-webapps/pull/232/files#diff-fccc52c233bf2632d2635c950bf8e88dR5

PR #233, opened 2018-05-16
I568 basic

littledice:
This is a basic implementation of https://react-table.js.org/#/story/readme with its default sorting and filtering. Behind a switch so can be merged in safely. To view, append any query string to the homepage.

Next PR will remove the redux sorting/filtering and use the table's custom filtering option to add more sophisticated field level filtering on the table.


app/src/main/report/components/ReportsList/ReportsListComponent.tsx
```diff
@@ -1,16 +1,24 @@
...
-    <ul className="list-unstyled">
-        {props.reports.map((report: Report) => (
-            <ReportListItem key={report.name} {...report} />
-        ))}
-    </ul>
-);
\ No newline at end of file
+export const ReportsListComponent: React.StatelessComponent<ReportsListComponentProps>
+    = (props: ReportsListComponentProps) => {
+    if (location.search.length) {
```
littledice: Just add any query string to view the new table instead of the old list.

PR #234, opened 2018-05-16
I568 customfiltering

littledice:
But first merge other 2 PRs for a cleaner diff


app/config/reporting_api_version
```diff
@@ -1 +1 @@
-master
\ No newline at end of file
+i1752
```
MartinEden: This can return to master

app/src/main/report/components/ReportsList/ReportListColumns/VersionColumn.tsx
```diff
@@ -15,6 +16,15 @@ export const latestVersionAccessorFunction = (data: Report): LatestVersion => {
     return {version: data.latest_version, date: new Date(data.updated_on)}
 };
 
+export const versionFilterMethod = (filter: FilterGeneric<VersionFilterValue>, row: ReportRowProps) => {
+    const lastUpdatedDate = row.latest_version.date;
+    const lastVersionId = row.latest_version.version;
+
+    return lastUpdatedDate <= filter.value.end
+        && lastUpdatedDate >= filter.value.start &&
+        lastVersionId.indexOf(filter.value.versionId) > -1;
```
MartinEden: Case insensitive, in case people use capital letters?

app/src/main/report/components/ReportsList/ReportListColumns/LatestVersionFilter.tsx
```diff
@@ -0,0 +1,62 @@
...
+    const onEndDateChange = (date: Date) => {
+        props.onChange({
+            end: date,
+            start: value.start,
+            versionId: value.versionId
+        })
+    };
+
+    return <div>
+        <input type={"text"} className={"form-control mb-1 "}
```
MartinEden: I feel like searching by ID might be less common than by date range, so maybe swap the order of these two? I see that this keeps it more in line with other columns, but I still think it might be nicer to have the more user friendly one first.

PR #235, opened 2018-05-16
remove old code and switch

littledice:
The most satisfying kind of PR!


app/src/test/report/reducers/ReportsReducersTests.ts
```diff
@@ -1,67 +0,0 @@
-import { expect } from "chai";
```
littledice: For some reason this class was a duplicate of the 1 below

littledice:
Ok, I've added a title and made the sorting hint into helper text. I'm not too sure about changing font sizes/weights in the table, my feeling is that the blue link text makes the name stand out very clearly already, and actually people are as likely to be searching based on metadata fields as report name!

PR #244, opened 2018-05-24
I1737 create user form

littledice:
This PR migrates the user creation form to `reduxForm.`


app/src/main/admin/actions/usersActionCreators.ts
```diff
@@ -16,4 +16,30 @@ export const usersActionCreators = {
...
         }
     },
+
+    createUser(name: string, email: string, username: string) {
+        return async (dispatch: Dispatch<AdminAppState>, getState: () => AdminAppState) => {
+            const result = await (new UsersService(dispatch, getState)).createUser(name, email, username);
+            if (result && result.errors) {
+                dispatch({
+                    type: UsersTypes.SET_CREATE_USER_ERROR,
+                    error: result.errors[0].message
```
MartinEden: Would be better to handle all errors coming back from the API, not just the first one.

app/src/main/admin/components/Users/Create/CreateUserForm.tsx
```diff
@@ -1,55 +1,112 @@
...
+import {ReduxFormValidationError} from "../../../../shared/components/ReduxForm/ReduxFormValidationError";
+import {AdminAppState} from "../../../reducers/adminAppReducers";
+import {usersActionCreators} from "../../../actions/usersActionCreators";
 
-export class CreateUserFormComponent extends React.Component<ReformProps, undefined> {
-    constructor() {
-        super();
-        this.changeName = this.changeName.bind(this);
+export interface CreateUserFormProps {
+    handleSubmit: (F: Function) => any,
```
MartinEden: Remind me again what's the deal with this thing? And maybe add a comment?

MartinEden:
Looks good. I get a weird thing where if I leave a field blank it just pops up "is required" or "is invalid" - missing the field name.

app/src/main/admin/components/Users/Create/CreateUserForm.tsx
littledice: Seems redux form adds this handler function and wants you to use it on form submission. Not sure what would happen if you just skip this and call our custom `submit` function directly. I might try that actually and see.

PR #245, opened 2018-05-29
I1737 createusersection

littledice:
Small PR to finish off the user creation migration to redux. First merge https://github.com/vimc/montagu-webapps/pull/244


app/package.json
```diff
@@ -81,6 +81,7 @@
     "@types/react-addons-test-utils": "^0.14.18",
     "@types/react-dom": "^15.5.0",
     "@types/react-redux": "^5.0.14",
+    "@types/react-router": "4.0.25",
```
littledice: The latest version of these typings (which are a dependency of the react typings) seem to have a bug. See failing TC build on master. So have pinned this version here.

PR #251, opened 2018-06-13
I1845 - User details component

littledice:
Including the interface to add/remove roles. Also deleted a fair bit of orphaned altjs code (the `UserStore` and `ModellingGroupStore` and associated tests)


littledice:
I can't replicate this. The logic to refresh the user is the same for adding and removing roles and I've factored it out into a private method to make this clearer. Could you try again and confirm whether you're still seeing this behaviour?

PR #252, opened 2018-06-14
i148

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-148

Very simple touchstone page that just lists basic metadata.


app/src/test/admin/actions/AdminTouchstoneActionTests.ts
```diff
@@ -0,0 +1,18 @@
...
+
+describe("Admin touchstone action tests", () => {
+    const sandbox = new Sandbox();
+    afterEach(() => sandbox.restore() );
+
+    it("gets all touchstones", (done: DoneCallback) => {
+        verifyActionThatCallsServiceAndReturnsResult(done, {
+            mockServices: () => sandbox.stubService(TouchstonesService.prototype, "getAllTouchstones"),
+            callActionCreator: () => adminTouchstoneActionCreators.getAllTouchstones(),
+            expectTheseActionTypes: [TouchstoneTypes.ALL_TOUCHSTONES_FETCHED]
```
littledice: I think it would be nice if this had the option of also verifying the data that is dispatched with the action to make it a bit more flexible; e.g. suppose a service returns "OK" and a notification is dispatched. But maybe that's a different helper method to keep this one simple. Just a thought.

PR #254, opened 2018-06-15
I1841 create new modelling group

littledice:
This PR is for the service, action creator and reducer. Components to follow.


app/src/test/ActionCreatorTestHelpers.ts
```diff
@@ -1,12 +1,33 @@
...
+    expectTheseActions: any[];
+}
+
 export interface VerifyActionThatCallsServiceAndReturnsResultProperties {
     mockServices: () => void;
     callActionCreator: () => void;
     expectTheseActionTypes: any[];
 }
 
+export function verifyActionThatCallsService(
```
littledice: I'm on board with introducing some more test helpers, it is nicer, you're right. I added a more general helper that allows you to check the full actions, not just the action types because in the case I was testing, the data returned by the service isn't returned with the action.

app/config/api_version
```diff
@@ -1 +1 @@
-master
+i1840
```
MartinEden: This can return to master now, I think.

littledice:
That test failure is on another branch (this one's `i1841_service`) so I'll go ahead and merge this after returning the API to master

MartinEden:
Oh sorry. Go for it!

PR #256, opened 2018-06-18
i1866: Handle compressed tokens

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-1866

These must be merged in iff merging https://github.com/vimc/montagu-api/pull/274.

`api_version` will also need returning to master in that situation.


app/scripts/cli.sh
```diff
@@ -4,5 +4,4 @@ api_version=$(<$here/../config/api_version)
 
 # Use this once you have run the API with ./run-apis.sh
 image=docker.montagu.dide.ic.ac.uk:5000/montagu-cli:$api_version
-docker pull $image
```
MartinEden: This is not really necessary, and it slows the various scripts down a bit.

PR #257, opened 2018-06-19
I1841 - add new group form

littledice:
Form component and show/hide form section


app/src/main/admin/components/ModellingGroups/Create/CreateModellingGroupForm.tsx
```diff
@@ -0,0 +1,135 @@
...
+                            label={"institution"}
+                            component={ReduxFormField}
+                            type="text"
+                            validate={[validations.required]}
+                            onChange={(e) => this.onInsititutionChange(e)}
+                        />
+                    </td>
+                </tr>
+                <tr>
+                    <td>Principal investigator</td>
```
MartinEden: Is that what that stands for? I wasn't sure!

littledice:
It actually doesn't handle nulls, rather it only gets called when both the inputs are non null. But I guess it wouldn't hurt to change it to handle nulls too.

littledice:
Oops, just overlooked that. Added some tests.

PR #260, opened 2018-06-19
I1839 - actions and service to get responsibilities for touchstone

littledice:


app/src/main/contrib/actions/contribTouchstonesActionCreators.ts
```diff
@@ -21,12 +21,7 @@ export const contribTouchstonesActionCreators = {
     setCurrentTouchstoneVersion(touchstoneVersionId: string) {
         return (dispatch: Dispatch<ContribAppState>, getState: () => ContribAppState) => {
             const touchstones = getState().touchstones.touchstones;
-            const versions = flatMap(touchstones, x => x.versions);
```
littledice: Factored out this logic into a shared action creator, so that it can be re-used (re-use will appear in the next PR though!)

PR #261, opened 2018-06-20
I1839

littledice:
This creates a page that just lists all responsibilities in a touchstone, grouped by responsibility set. We'll need to think more about layout/design but it's s start?


app/src/test/mocks/mockModels.ts
```diff
@@ -100,8 +100,8 @@ export function mockTouchstoneVersion(properties?: Partial<TouchstoneVersion>):
     return Object.assign(template, properties);
 }
 
-export function mockResponsibilitySet(properties?: any,
```
littledice: This was wrongly named.

app/src/main/admin/components/AdminRouter.tsx
```diff
@@ -30,8 +31,9 @@ export const AdminRouter : React.StatelessComponent<AdminRouterProps> = (props:
         <Route exact path="/modelling-groups/:groupId/" component={ModellingGroupDetailsPage}/>
         <Route exact path="/modelling-groups/:groupId/admin/" component={ModellingGroupMembersPage}/>
         <Route exact path="/touchstones/" component={TouchstoneListPage}/>
+        <Route exact path="/touchstones/:touchstoneVersionId/" component={TouchstoneVersionPage}/>
```
MartinEden: So, in VIMC-1885, I have added `/touchstones/:touchstoneId`. Maybe the URL scheme for versions of touchstone should be: `/touchstones/:touchstoneId/:versionId`, e.g. `/touchstones/201810gavi/3/`?

littledice: ah yes, makes sense

PR #263, opened 2018-06-22
i1885: Add touchstone details page 

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-1885


app/src/main/shared/actions/touchstoneActionCreators.ts
```diff
@@ -1,17 +1,25 @@
-import {SetCurrentTouchstoneVersion, TouchstoneTypes} from "../actionTypes/TouchstonesTypes";
+import {SetCurrentTouchstone, SetCurrentTouchstoneVersion, TouchstoneTypes} from "../actionTypes/TouchstonesTypes";
 import {flatMap} from "../ArrayHelpers";
 import {Touchstone} from "../models/Generated";
 
 export const touchstonesActionCreators = {
+    setCurrentTouchstone(touchstoneId: string, touchstones: Touchstone[]) {
+        const currentTouchstone = touchstones.find(t => t.id === touchstoneId);
+        if (!currentTouchstone) {
+            throw Error("Unknown touchstone: " + touchstoneId);
```
littledice: This is obviously useful for debugging, but from the user's point of view if they navigate to a url with a non-existent touchstone id, they just get a loading spinner. Intuitively I think they should see a page not found error in that case? 

Not in scope for this ticket, but it'd be nice if this type of error could direct you to the page not found message.

MartinEden:
I've updated this now that your responsibilities page work is in.

As to error messages on missing objects: I agree, although I don't have any detailed thoughts at the moment.

PR #264, opened 2018-06-22
I1842 - create touchstone form

littledice:
Although I haven't written explicit tests of `montaguForm` it is implicitly tested via `CreateUserFormTests`, `CreateTouchstoneFormTests` and `CreateGroupFormTests` so I'm pretty happy that it's under test.


app/src/main/admin/components/Touchstones/List/TouchstoneListPage.tsx
```diff
@@ -20,6 +21,7 @@ export class TouchstoneListPageComponent extends React.Component<PageProperties<
         return <div>
             <AdminPageHeader />
             <PageArticle title={TouchstoneListPageComponent.title}>
+                <CreateTouchstoneForm/>
```
littledice: Should remove this before merging, as it's not wired up.

app/src/main/shared/components/ReduxForm/MontaguForm.tsx
```diff
@@ -0,0 +1,31 @@
+import * as React from "react";
```
littledice: What do you think of this HOC? Have applied it to the create user and create modelling groups forms as well as the new touchstone form. Do you think it makes the forms simpler to write or is it unnecessary ?

app/src/main/admin/components/Touchstones/List/TouchstoneListPage.tsx
MartinEden: Could you instead use a flag (e.g. in development settings, but not in production) so hide it?

app/src/main/admin/actions/usersActionCreators.ts
```diff
@@ -24,7 +25,8 @@ export const usersActionCreators = {
         }
     },
 
-    createUser(name: string, email: string, username: string) {
+    createUser(values: CreateUserFormFields) {
+        const {name, email, username} = values;
```
littledice: I changed the signature of this method because named parameters are less error prone than positional.

app/src/test/admin/components/Touchstones/List/TouchstoneListItemTests.tsx
```diff
@@ -13,6 +13,8 @@ describe("TouchstoneListItem", () => {
         const rendered = shallow(<TouchstoneListItem {...t}/>);
         const cells = rendered.find("td");
 
+        expect(cells.length).to.eq(4);
+
         const descriptionCell = cells.at(1);
         expect(descriptionCell.find(InternalLink).dive().text()).to.eql("desc1");
         expect(descriptionCell.find(InternalLink).prop("href")).to.eql("/touchstones/t1/");
```
MartinEden: Would be good to have a test that it doesn't error when there are no versions in the touchstone, given that was a thing you fixed in this change.

PR #266, opened 2018-07-12
save compressed token and pass it back as bearer token

littledice:


app/src/@types/globals.d.ts
```diff
@@ -15,5 +15,6 @@ declare module 'react-day-picker/DayPickerInput';
 declare module 'react-day-picker/moment';
 declare module 'reselect';
 declare module 'moment';
+declare module 'pako'
```
littledice: not sure why this wasn't throwing up compile errors before :woman_shrugging:

PR #269, opened 2018-07-16
log out if token is invalid

littledice:


app/src/main/shared/actions/authActionCreators.ts
```diff
@@ -37,11 +37,17 @@ export const authActionCreators = {
         return (dispatch: Dispatch<any>) => {
             const token = localStorageHandler.get("accessToken");
             if (token) {
-                const inflated = jwtTokenAuth.inflateToken(token);
+                let inflated = token;
+                try {
+                    inflated = jwtTokenAuth.inflateToken(token);
+                }
+                catch(e){
+                    console.log("Access token has expired or is otherwise invalid: Logging out.");
```
MartinEden: I thought the idea was to display a message to the user? That was why I was suggesting you could just reuse the existing message. If you're just logging it then I would be more specific and technical.

app/src/main/shared/actions/authActionCreators.ts
```diff
@@ -37,11 +37,17 @@ export const authActionCreators = {
         return (dispatch: Dispatch<any>) => {
             const token = localStorageHandler.get("accessToken");
             if (token) {
-                const inflated = jwtTokenAuth.inflateToken(token);
+                let inflated = token;
```
MartinEden: Shouldn't this just be something like `let inflated: string = null;`

app/src/main/shared/actions/authActionCreators.ts
littledice: Well the point was for it not to break everything. We could log a message to the user and a more technical message to the console but we don't seem to be doing that in the other error path of this method. In fact the other error path seems to fully obscure what's going on - `jwtTokenAuth.decode` contains a try/catch which console logs a message and then returns a token with a null expiry date which hits the `isExpired` check in this method!

MartinEden: Fair enough. In that case just make the message a bit more specific, to say that decompressing is what failed, rather than decoding?

littledice: I think it's not neccessary to inform the user what's going on. Logging them out silently is the behaviour when a token in storage has expired, so as far as they're concerned this is the same thing.

PR #271, opened 2018-07-19
I1955 onetimelinkcontext component

littledice:
Quite a few of these file changes are a rename (renamed directory actions -> actionCreators) and a couple of fixed typings that aren't anything to do with the substantive change here, should have done them on another branch really but hopefully they don't clutter this PR too much :crossed_fingers: 

The 2 main things to note about this refactor:
1. Treating one time tokens as strings and just passing them around instead of decoding them
2. No caching. So every time a component involving a one time link mounts, it will fetch a new token. This keeps the code simple - to refresh a token, you just call `fetchToken(url)` again. It does mean more API calls though. If you think we should re-introduce caching, new style, I'll do that in another PR.

One more PR after this to delete all orphaned `altjs` code.


app/src/main/report/components/OneTimeLinkContext.tsx
```diff
@@ -1,73 +1,60 @@
...
+import {oneTimeTokenActionCreators} from "../actionCreators/oneTimeTokenActionCreators";
+import {buildReportingURL} from "../services/AbstractReportLocalService";
 
 interface PublicProps {
     href: string;
     className?: string;
 }
 
-interface Props extends PublicProps {
-    token: OneTimeToken;
```
littledice: So one thing to note is that I'm not using this `OnetimeToken` type anywhere, but just treating the token as a string to be retrieved and passed back to the API, without decoding/looking inside. I couldn't see a good reason to be doing the latter, but feel free to point out if I'm wrong about this?

app/src/test/mocks/mockStore.ts
```diff
@@ -11,14 +12,14 @@ export function createMockAdminStore(initialState?: RecursivePartial<AdminAppSta
...
-export function createMockReportStore(initialState?: RecursivePartial<ReportAppState>) {
+export function createMockReportStore(initialState?: RecursivePartial<ReportAppState>): MockStore<ReportAppState> {
     return createMockStore(initialState);
 }
 
 /** This method is deprecated. Use the more specific versions above to get compile-time
  * errors
  */
-export const createMockStore = (initialState: any = {}) => {
+export const createMockStore = (initialState: any = {}): MockStore<any> => {
```
littledice: Opportunistic type fix

app/src/test/report/actions/onetimeTokenActionCreatorsTests.ts
```diff
@@ -0,0 +1,34 @@
...
+
+import {Sandbox} from "../../Sandbox";
+import {oneTimeTokenActionCreators} from "../../../main/report/actionCreators/oneTimeTokenActionCreators";
+import {OneTimeTokenService} from "../../../main/report/services/OneTimeTokenService";
+import {createMockReportStore} from "../../mocks/mockStore";
+import {ReportAppState} from "../../../main/report/reducers/reportAppReducers";
+import {MockStore} from "redux-mock-store";
+import {OnetimeTokenActionType} from "../../../main/report/actionTypes/OnetimeTokenActions";
+
+describe("User action creators", () => {
```
MartinEden: Wrong test name

littledice: Whoops, thanks!

littledice:
Typo fixed! And yep, have tested manually locally.

PR #272, opened 2018-07-19
I1955 clean up orphaned code

littledice:
The reportle code is now 100% `altjs`-free.


app/src/integrationTests/ReportIntegrationTests.tsx
```diff
@@ -90,13 +82,6 @@ class ReportIntegrationTests extends IntegrationTestSuite {
             expect(versionDetails.name).is.equal("minimal");
         });
 
-        it("fetches report one time token", async () => {
-            const versions = await (new ReportsService(this.store.dispatch, this.store.getState)).getReportVersions("minimal");
```
littledice: This was testing the old `altjs` source; the same test for the new service is directly below.

app/src/integrationTests/ReportIntegrationTests.tsx
```diff
@@ -109,51 +94,57 @@ class ReportIntegrationTests extends IntegrationTestSuite {
         it("downloads artefact", async () => {
             const artefact = mockArtefact({filenames: ["all.csv", "all.png"], description: "all things"});
             const versions = await (new ReportsService(this.store.dispatch, this.store.getState)).getReportVersions("multi-artefact");
-            const rendered = shallow(<ArtefactItem report={"multi-artefact"} version={versions[0]}
-                                                   artefact={artefact}/>);
+            const rendered = sandbox.mount(<Provider store={this.store}><ArtefactItem report={"multi-artefact"} version={versions[0]}
```
littledice: Testing the full mounted component (so implicitly also testing the `OneTimeLinkContext` here, which we weren't doing before.)

app/src/test/mocks/mockRemote.ts
```diff
@@ -57,7 +57,7 @@ export function resetFetcher() {
 }
 
 export function mockFetcher(promise: Promise<Response>, reportingPromise?: Promise<Response>) {
-    fetcher.fetcher = new ReportingFetcher();
+    fetcher.fetcher = new AdminFetcher()
```
littledice: I've deleted the `ReportingFetcher` altogether, but admin still uses the old fetcher, so we can't get rid of this quite yet.

PR #274, opened 2018-07-23
I1969 - remove `Responsibilities` and replace with `ResponsibilitySet`

littledice:
bring portals in line with API changes


app/src/test/mocks/mockEndpoints.ts
```diff
@@ -1,61 +0,0 @@
-import {APIType, FakeEndpoint} from "./mockMultipleEndpoints";
```
littledice: This was just an unused file

PR #276, opened 2018-07-23
update expectations UI and tests

littledice:
Ready for review. UI based on feedback from Katy and Margaret

### DIFF
--- a/app/src/main/contrib/components/Responsibilities/Expectations/ExpectationsDescription.tsx
+++ b/app/src/main/contrib/components/Responsibilities/Expectations/ExpectationsDescription.tsx
@@ -1,5 +1,5 @@
 import * as React from "react";
-import {ExpectationMapping, NumberRange} from "../../../../shared/models/Generated";
+import {CohortRestriction, ExpectationMapping, NumberRange} from "../../../../shared/models/Generated";
 import {CountriesList} from "./CountriesList";
 
 export class ExpectationsDescription extends React.PureComponent<ExpectationMapping> {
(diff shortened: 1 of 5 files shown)


app/src/main/contrib/components/Responsibilities/Expectations/ExpectationsDescription.tsx
```diff
@@ -1,15 +1,68 @@
 import * as React from "react";
-import {ExpectationMapping} from "../../../../shared/models/Generated";
+import {Country, ExpectationMapping, NumberRange} from "../../../../shared/models/Generated";
+import {Popover, PopoverBody} from "reactstrap";
+
+interface State {
+    showCountries: boolean;
+}
```
MartinEden: I don't having interfaces just called `Props` and `State`, so long as they are in a file with just one component. It gets a bit confusing here. I would suggest you move `CountriesList` and its bits to another file.

app/src/main/contrib/components/Responsibilities/Expectations/ExpectationsDescription.tsx
```diff
@@ -1,15 +1,68 @@
...
+export class ExpectationsDescription extends React.Component<ExpectationMapping, State> {
 
-export class ExpectationsDescription extends React.PureComponent<ExpectationMapping> {
     render(): JSX.Element {
+
+        const expectation = this.props.expectation;
+        const numCountries = expectation.countries.length;
+        const numYears = expectation.years.maximum_inclusive - expectation.years.minimum_inclusive + 1;
+        const numAges = expectation.ages.maximum_inclusive - expectation.ages.minimum_inclusive + 1;
+        const numCohorts = expectation.cohorts.maximum_birth_year - expectation.cohorts.minimum_birth_year + 1;
```
MartinEden: What if one (or both) of the cohorts values is null? I'm also not sure we need to list the number of cohorts - it makes less sense than the others. See my suggested output in another comment below.

app/src/main/contrib/components/Responsibilities/Expectations/ExpectationsDescription.tsx
```diff
@@ -1,15 +1,68 @@
...
             <div>
-                You need to do the following:
+                Expecting data for:
                 <ul>
-                    <li>These years</li>
-                    <li>These ages</li>
+                    <li id={"years"}>{numYears} years: {rangeAsString(expectation.years)}</li>
+                    <li id={"ages"}>{numAges} ages: {rangeAsString(expectation.ages)}</li>
+                    {expectation.cohorts.minimum_birth_year &&
+                    <li id={"cohorts"}>{numCohorts} cohorts: {expectation.cohorts.minimum_birth_year} - {expectation.cohorts.maximum_birth_year}</li>}
```
MartinEden: Talk to Tini, but I think these cohorts need to be phrased with more explanation than just the simple age and year restrictions. Ideally we want this to be clear to prospective modellers without them reading lots of explanations. I was thinking something like these four (based on which values are null):

* "No cohort restrictions"
* "Cohorts: Only people born after 2000"
* "Cohorts: Only people born before 2050"
* "Cohorts: Only people born after 2000 and before 2050"

app/src/main/contrib/components/Responsibilities/Expectations/ExpectationsDescription.tsx
```diff
@@ -1,15 +1,68 @@
...
+                    <li id={"countries"}>{numCountries} countries: <CountriesList targetKey={`countries-${expectation.id}`}
+                                                                                  countries={expectation.countries} /></li>
+                    <li id={"outcomes"}>{numOutcomes} outcomes: {expectation.outcomes.join(", ")}</li>
```
MartinEden: I had been imagining handling the outcomes a bit differently. Because the others are the dimensions we want to capture outcomes across, whereas the outcomes are what we actually want to capture. Maybe talk to Tini, but I had been thinking along the lines of:

```
We are expecting these 7 outcomes: [table including full outcome descriptions]

We are expecting these outcomes to be provided for all combinations of:
* 98 countries...
* 100 ages...
* etc.
```

littledice:
Tagging @tinigarske to get her input on the phrasing of these thing ^

app/src/main/contrib/components/Responsibilities/Expectations/ExpectationsDescription.tsx
littledice: I'm inclined to agree with all your suggestions, I'll go ahead and make these changes

littledice: Updated to be split out outcomes from rows, as suggested. We can't display the full outcome descriptions without further API work though, as we only return the name.

PR #277, opened 2018-07-24
migrate contrib page action creators to new style

littledice:
This refactors the top-level page in the Contrib portal, `ChooseActionPage`, to use the new `AbstractPageActionCreators` class, and also adds a new HOC that can be used to hook up an instance of the `AbstractPageActionCreators` class to a page component by automatically calling the former's `onLoad` method when the component mounts.

### DIFF
--- a/app/src/test/ActionCreatorTestHelpers.ts
+++ b/app/src/test/ActionCreatorTestHelpers.ts
@@ -40,7 +40,7 @@ export function verifyActionThatCallsServiceAndReturnsResult(
     store.dispatch(props.callActionCreator());
     setTimeout(() => {
         const actions = store.getActions();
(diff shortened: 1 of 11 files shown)


app/src/test/ActionCreatorTestHelpers.ts
```diff
@@ -40,7 +40,7 @@ export function verifyActionThatCallsServiceAndReturnsResult(
     store.dispatch(props.callActionCreator());
     setTimeout(() => {
         const actions = store.getActions();
-        expect(props.expectTheseActionTypes).to.eql(actions.map((x: any) => x.type));
```
littledice: Just swapped the order of this assertion so the fail message makes more sense.

littledice:
Precisely :+1:

PR #289, opened 2018-07-30
i1853: Refactor admin forgotten password

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-1853


app/src/main/admin/components/AdminForgottenPasswordPage.tsx
```diff
@@ -1,28 +1,34 @@
...
+            urlFragment: "forgotten-password/"
+        }
     }
 
     render(): JSX.Element {
-        return <Page page={this}>
-            <ForgottenPasswordForm />
-        </Page>;
+        return <div>
+            <AdminPageHeader/>
```
littledice: Am I right in thinking that now everything is redux we can move this `AdminpageHeader` component to the router (similarly for the other portals) and then we could have a single `ForgottenPasswordPageComponent` that all 3 portals route to? I think the only difference between these 3 pages is the header...

app/src/main/admin/components/AdminRouter.tsx
```diff
@@ -27,7 +26,7 @@ interface AdminRouterProps {
 export const AdminRouter : React.StatelessComponent<AdminRouterProps> = (props: AdminRouterProps) => {
 
     const loggedIn = <Switch>
-        <Route exact path="/" component={MainMenu}/>
+        <Route exact path="/" component={MainMenuNew}/>
```
littledice: We should also go through and rename all the `xNew` components to `x`, but maybe there's another PR for this?

MartinEden: Yup, that's the last PR

app/src/main/admin/components/AdminForgottenPasswordPage.tsx
MartinEden: Feel like this ticket has enough PRs. So here's a new ticket for either of us to work on: https://vimc.myjetbrains.com/youtrack/issue/VIMC-2020

PR #292, opened 2018-07-30
i2020: Lift `AdminPageHeader` up to router

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-2020


app/src/main/admin/components/AdminRouter.tsx
```diff
@@ -48,6 +50,7 @@ export const AdminRouter : React.StatelessComponent<AdminRouterProps> = (props:
     const routes = props.loggedIn ? loggedIn : notLoggedIn;
 
     return <ConnectedRouter history={props.history}>
+        <PageHeaderNew siteTitle={"Admin portal"} logo={logo} />
```
littledice: Couple of test failures; I think it might be complaining that the `Router` should only have one child... so maybe it needs to be
```
<div>
<Header />
<Router>
...
</Router>
</div>
```

littledice: Couple of test failures; I think it might be complaining that the `Router` should only have one child... so maybe it needs to be
```
<div>
<Header />
<Router>
...
</Router>
</div>
```

PR #295, opened 2018-07-30
i2023: Autofill email

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-2023

This isn't really part of the original ticket, but I couldn't resist fixing this while I was here, as it has bugged me for ages. Now if you click "Forgotten password", it will carry over your email from the login form, saving you keystrokes.

### DIFF
--- a/app/src/main/shared/components/Login/LoginForm.tsx
+++ b/app/src/main/shared/components/Login/LoginForm.tsx
@@ -1,37 +1,43 @@
 import * as React from "react";
-import { reduxForm, Field } from "redux-form";
-import { connect } from "react-redux";
-import { compose } from "recompose";
-import { Dispatch } from "redux";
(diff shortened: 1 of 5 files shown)


app/src/main/shared/components/Login/LoginForm.tsx
```diff
@@ -1,37 +1,43 @@
...
-export interface LoginFormFields{
+export interface LoginFormFields {
     email: string;
     password: string;
 }
 
 export class LoginFormComponent extends React.Component<LoginFormProps, undefined> {
     renderField(data: InputFieldProps) {
-        const { input, label, type, meta: { touched,  error } } = data;
+        const {input, label, type, meta: {touched, error}} = data;
```
littledice: I think this whole method can be replaced by the `ReduxFormInput` component

PR #296, opened 2018-07-30
refactor contrib page action creators

littledice:
There are 2 things this PR doesn't do.

1. refactor tests (much) - the tests for the page action creators can be simplified now, but I'll do a separate PR as this one is already quite big.
2. convert pages that don't currently have any action creators. E.g. the guidance pages. This means that they manage their breadcrumbs differently, so perhaps we should convert them all as well, it might be easier to reason about if all pages are the same. But again, if you agree, I'll do this in another PR.


app/src/test/contrib/components/ChooseAction/ChooseActionPageTests.tsx
```diff
@@ -14,24 +14,18 @@ import {chooseActionPageActionCreators} from "../../../../main/contrib/actions/p
 import {ChooseActionContent} from "../../../../main/contrib/components/ChooseAction/ChooseActionContent";
 import {mockMatch} from "../../../mocks/mocks";
 
-describe("Choose Action Page Component", () => {
+describe("Choose Action Page", () => {
 
     const sandbox = new Sandbox();
     afterEach(() => sandbox.restore());
 
-    it("renders component on connect level", () => {
```
littledice: I deleted this as we already check that the page action creator's `onLoad` method is called in the next test. Similarly for the other test suites.

PR #300, opened 2018-08-08
i2042: Refactor demographic and contrib

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-2042

This PR removes the onetime button blocker, and just implements a simple onetime button. I'll add the temporarily disabling the button logic back in the next PR.


app/src/main/shared/components/OneTimeLinkContext.tsx
```diff
@@ -58,10 +66,19 @@ export function OneTimeLinkContext(WrappedComponent: ComponentConstructor<OneTim
                 className={this.props.className}
                 href={href}
                 service={this.props.service}
+                enabled={this.getEnabled()}
                 refreshToken={() => this.props.refreshToken(this.props.href, service)}
                 children={this.props.children}/>;
         }
 
+        private getEnabled(): boolean {
+            if (isNullOrUndefined(this.props.enabled)) {
```
littledice: Given this is a boolean property and we want it to be true when undefined, does it not make sense to invert and have a an optional 'disabled' property instead? Then the undefined state will simply evaluate to false.

MartinEden: Maybe. I really don't like "disabled" properties. And I think it will be more difficult to reason once you bring in the temporarily disabled logic in the next PR. However, I was thinking I could maybe provide the default prop in the `mapStateToProps` step, and then the component wouldn't have to worry about it. How would you feel about that?

littledice: I mean, that's exactly what's going on here, you're having to invert logic ;) But sure, lets re-evaluate after next PR.

PR #301, opened 2018-08-08
Weirdbug

MartinEden:
So it looks like we needed one more `dive`. I'm not sure why this changed.

However, it general I think it's pretty terrible all this "do we want 4 dives or 5 or 6" nonsense. Could we write some kind of `.diveUntilYouFindSomeHTML()` function? That would handle a bunch of cases, although sadly not all. Maybe `diveUntilIsPresent(SomeComponent)`?


littledice:
I'm soooo dissatisfied with this - how can this have gone from passing to not passing with no code changes AND no other tests affected?!!?!? :cry: And I agree, all the diving is horrible. Found some discussion of this issue here and linked issues: https://github.com/airbnb/enzyme/issues/1138

PR #304, opened 2018-08-10
i2056: Change uploads in contrib portal to use new-style one time tokens

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-2056

Need to return to API master after https://github.com/vimc/montagu-api/pull/302 has been merged.

I noticed we had a manually defined type `EstimatesCreateBurdenData` for the data we send to create a burden estimate set, but this is of course unnecessary, as there's a class `CreateBurdenEstimateSet` in the shared webmodels code. So I switched over to that here. This type was very slightly different, as it requires a `model_run_parameter_set` value, but this is null in the case of central estimates, which is what we're dealing with here, so simple to add null in a few places.


app/src/main/contrib/components/Responsibilities/BurdenEstimates/UploadBurdenEstimatesContent.tsx
```diff
@@ -91,6 +89,6 @@ export const mapStateToProps = (state: ContribAppState): Partial<UploadBurdenEst
 };
 
 export const UploadBurdenEstimatesContent = compose(
-    connect(mapStateToProps, ),
+    connect(mapStateToProps, (dispatch, props) => props),
```
MartinEden: Otherwise it adds a `dispatch` function to the props, which we don't need and it muddies the tests.

PR #305, opened 2018-08-13
I2042 alex suggestions

littledice:
Removes `enabled` prop passed to `OnetimeLinkContext` and introduces `loading` prop generated by the HOC and passed to the wrapped component.

### DIFF
--- a/app/src/main/contrib/components/Responsibilities/Demographics/DownloadDemographicsContent.tsx
+++ b/app/src/main/contrib/components/Responsibilities/Demographics/DownloadDemographicsContent.tsx
@@ -22,7 +22,10 @@ export class DownloadDemographicsContentComponent extends React.Component<Downlo
         const {touchstone, selectedDataSet, selectedFormat, selectedGender} = this.props;
         let url: string = null;
         if (canDownload) {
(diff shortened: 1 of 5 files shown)


app/src/main/contrib/components/Responsibilities/Demographics/DownloadDemographicsContent.tsx
```diff
@@ -22,7 +22,10 @@ export class DownloadDemographicsContentComponent extends React.Component<Downlo
         const {touchstone, selectedDataSet, selectedFormat, selectedGender} = this.props;
         let url: string = null;
         if (canDownload) {
-            url = `/touchstones/${touchstone.id}/demographics/${selectedDataSet.source}/${selectedDataSet.id}/?format=${selectedFormat}&gender=${selectedGender}`;
+            url = `/touchstones/${touchstone.id}/demographics/${selectedDataSet.source}/${selectedDataSet.id}/?format=${selectedFormat}`
```
littledice: Noticed this bug had crept in with the last merged refactor - if a gender was selected and then a non-gendered dataset selected, the file was coming back as e.g. `...-female.csv`. Perhaps we should also change the API which naively appends the query parameter to the end of the filename without checking whether it's relevant.

PR #307, opened 2018-08-14
introduce adminpage hoc and add 'title' method to page action creators

littledice:
Introduced an `AdminPage` hoc, just as we have a `ContribPage` one, and migrated the `TouchstoneList` and `TouchstoneDetails` pages over to using it.


app/src/main/shared/actions/AbstractPageActionCreators.ts
```diff
@@ -6,6 +6,10 @@ export abstract class AbstractPageActionCreators<TState, TPageProps> {
 
     abstract parent: AbstractPageActionCreators<TState, any>;
 
+    title(state: TState): string {
```
littledice: I've not made this abstract so that it doesn't require us to change every inherited class.

MartinEden: Agreed :)

app/src/main/shared/components/Page.tsx
```diff
@@ -16,13 +16,19 @@ export function Page<TState, TLocationProps>(pageActionCreators: AbstractPageAct
         }
     };
 
+    const mapStateToProps = (state: TState): Partial<PageProperties<TLocationProps>> => {
+        return {
+            title: pac.title(state)
+        }
+    };
```
MartinEden: Is there a test for this bit?

littledice: Ah, not in the generic tests, good catch

PR #315, opened 2018-08-16
i2088: Disable wrapped one time link component when href is null

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-2088

I think this is the piece that got missed during the enable/loading refactor.


littledice:
Cool, nice easy fix :) I think now that you're doing the null check on `href` in `OnetimeLinkContext` we could also remove it from `FileDownloadInner` but doesn't really matter.

PR #328, opened 2018-08-20
I2091 -integration tests for components

littledice:


app/src/integrationTests/IntegrationTest.ts
```diff
@@ -69,8 +79,40 @@ export abstract class IntegrationTestSuite {
     }
 }
 
+export async function firstDownloadPromise(rendered: ReactWrapper) {
```
littledice: I moved these out to be globally available, but actually didn't end up using them. In any case, makes sense for them to be shared methods, so left them here.

PR #329, opened 2018-08-21
i2082 - add links to download coverage

littledice:
Show them as disabled with explanatory tooltip if user does not have global coverage reading permission.


MartinEden:
Waiting for TC.

PR #335, opened 2018-08-22
i2060: Remove dead code, including old style one time links

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-2060

In order to find the dead one time link code I used WebStorm's automated dead code detection tools, which found quite a bit more, so I've gone ahead and trimmed away anything we weren't using.


app/src/test/mocks/mocks.ts
```diff
@@ -21,20 +19,21 @@ export function mockMatch<P>(params?: P) :match<P> {
...
         goForward: () => {},
-        block: () => {},
-        listen: () => {},
-        createHref: () => {}
-    }, params);
+        block: (() => {}) as any,
+        listen: (() => {}) as any,
+        createHref: () => ""
+    };
+    return Object.assign(template, params);
```
MartinEden: The dead code inspector wanted to remove most of these properties because we weren't using them, so I made the method strongly typed.

PR #339, opened 2018-08-24
I2054 - show all report versions in main table

littledice:


app/src/main/report/components/ReportsList/ReportListTable.tsx
```diff
@@ -1,39 +1,42 @@
...
     name: string,
     author: string,
     requester: string,
-    published: boolean,
+    published?: boolean,
+
+    [key: string]: string | boolean | BasicVersionDetails | ReportRow[]
 
-    [key: string]: string | boolean | LatestVersion
+    _subRows?: ReportRow[]
```
MartinEden: Does this need an underscore in its name to work with the plugin? If not, it's quite out of keeping with our normal code style.

app/src/main/report/components/ReportsList/ReportListTable.tsx
```diff
@@ -54,6 +57,16 @@ export const TextFilter: FilterRender = (props: FilterProps<string>) => {
                   onChange={event => props.onChange(event.target.value)}/>
 };
 
+export const EmptyCell = () => {
+    return <span>&nbsp;</span>
+};
+
+export const NonEmptyCell = (row: ReportRowRenderProps) => {
```
MartinEden: Maybe call this one something like `CellWithValue`?

app/src/main/report/components/ReportsList/ReportListTable.tsx
littledice: Unfortunately I can't find another way of interacting with the sub rows, other than by this (presumably intended to be private) property.

littledice: Figured this out, you can actually specify a `subRowsKey` property in the component props to give this a nicer name :)

PR #348, opened 2018-11-14
I2337 - pinned reports

hillalex:
* adds pinned reports widget to homepage
* reports to pin currently determined by settings config file

### DIFF
--- a/app/src/main/report/components/ReportsList/PinnedReports.tsx
+++ b/app/src/main/report/components/ReportsList/PinnedReports.tsx
@@ -34,7 +34,7 @@ export class PinnedReportsComponent extends React.Component<PinnedReportsProps> 
             <div className={"row mb-5"}>
                 {this.props.reports.map((r: Report) =>
 
-                    <div className={"col-12 col-sm-6 col-lg-4"}>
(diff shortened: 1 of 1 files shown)


app/src/main/shared/styles/common.scss
```diff
@@ -89,5 +89,11 @@ select.form-control {
 }
 
 .download-icon {
-  margin-left: 15px;
+  margin-left: 10px;
+
+  svg {
```
littledice: Just an opportunistic fix - the styling on these icons was a bit screwed up.

app/src/main/shared/components/DownloadIcon.tsx
```diff
@@ -6,7 +6,7 @@ interface Props {
 // svg taken from the octicons library made available under the MIT license
 // https://github.com/primer/octicons/blob/master/LICENSE
 export const DownloadIcon = (props: Props) => {
-    return <svg viewBox="0 0 16 12" width="20" height="22"
+    return <svg viewBox="0 0 16 16" width="20"
```
littledice: opportunistic fix, see comment on .scss file below

PR #356, opened 2018-12-05
I2389

EmmaLRussell:


app/src/main/shared/styles/common.scss
```diff
@@ -97,3 +97,7 @@ select.form-control {
     margin-bottom: -1px;
   }
 }
+
+.render-whitespace {
+    white-space: pre-line;
```
littledice: nice, I never knew about this css property!

PR #357, opened 2018-12-06
removing instructions for stochastic runs as not required in 2018 tou…

EmmaLRussell:
…chstone


app/src/main/contrib/components/Responsibilities/Guidance/content/ResponsibilityGuidanceModelOutputsContentLatest.tsx
```diff
@@ -39,25 +39,6 @@ export class ResponsibilityGuidanceModelOutputsContentLatestComponent extends Re
                     </td>
                     <td>Montagu</td>
                 </tr>
-                <tr>
```
EmmaLRussell: This table now looks a bit silly as it only has one row left, but it is consistent with previous touchstones. Leave as is or reformat the info so it's just in a sentence rather than a table?

PR #359, opened 2018-12-07
I2405

EmmaLRussell:


app/src/main/report/actionCreators/reportPageActionCreators.ts
```diff
@@ -14,9 +14,14 @@ export const reportPageActionCreators = {
                 type: ReportTypeKeys.REPORT_VERSION_DETAILS_FETCHED,
                 data: null
             } as ReportVersionDetailssFetched);
+            dispatch({
```
littledice: I'm bit confused by this - why set the changelog to null here and then immediately reset it to data fetched by the API on line 24?

app/src/main/report/actionCreators/reportPageActionCreators.ts
```diff
@@ -14,9 +14,14 @@ export const reportPageActionCreators = {
...
+                type: ReportTypeKeys.REPORT_VERSION_CHANGELOG_FETCHED,
+                data: null
+            } as ReportVersionChangelogFetched);
 
             dispatch(reportActionCreators.getReportVersions(props.report));
             dispatch(reportActionCreators.getVersionDetails(props.report, props.version));
+            dispatch(reportActionCreators.getVersionChangelog(props.report, props.version));
```
littledice: At the moment this is going to throw an error if the user is not a report reviewer so this would cause an error message for all normal users visiting a report page. It should probably be dispatched by the `ReportChangelog` component itself, seeing as only reviewers will be given access to that.

app/src/main/report/services/ReportsService.ts
```diff
@@ -17,6 +17,11 @@ export class ReportsService extends AbstractLocalService {
             .get(`/reports/${reportId}/versions/${versionId}/`, "reporting");
     }
 
+    getVersionChangelog(reportId: string, versionId: string) {
```
littledice: Can we get an integration test for this?

app/src/main/report/actionCreators/reportPageActionCreators.ts
littledice: Ah ok, I see you're copying the logic with the version details. I'm not 100% sure why that's being done or whether it's necessary, but my guess is that it's so when you change from one report page to another, the previous report's content doesn't linger. This wouldn't be an issue with the changelog if you fetched it here. 

However see comment below about where I think the changelog should be fetched - in that case, this approach of first setting it to null might well be needed, to avoid stale content being shown when the component first mounts.

app/scripts/add-test-accounts-for-integration-tests.sh
```diff
@@ -24,3 +24,6 @@ $here/cli.sh addRole report.reviewer reports-reviewer
 $here/cli.sh addRole report.reviewer user-manager
 $here/cli.sh addRole report.reviewer admin
 $here/cli.sh addRole report.reviewer reports-reader report:html
```
EmmaLRussell: I wanted to make the report reviewer user be logged in for the report integration tests in order to see the changelog report versions. However fetching data from the report api appeared to fail unless the user was added to the modelling group. I have no idea why!!

PR #361, opened 2018-12-10
I2437

EmmaLRussell:


app/src/@types/Settings.d.ts
```diff
@@ -6,8 +6,9 @@ interface Settings {
     vimcEmail: string;
     slackUrl: string;
     isApplicantTouchstone: (id: string) => boolean;
-    isStochasticTouchstone: (id: string) => boolean;
     is2017Touchstone: (id: string) => boolean;
+    isVersionOfStochasticTouchstone: (touchstoneId: string) => boolean;
```
EmmaLRussell: Named this way partly to distinguish from old isStochasticTouchstone method, which should now be obsolete

app/src/main/contrib/components/Responsibilities/Overview/ResponsibilityOverviewDescription.tsx
```diff
@@ -95,15 +61,35 @@ export const ResponsibilityOverviewDescription = (props: ResponsibilityOverviewD
     }
 
     if (settings.isApplicantTouchstone(props.currentTouchstoneId)) {
```
EmmaLRussell: Old hardcoded stochastic touchstone no longer required. We could probably take out the JanuaryRfp touchstone logic too.

PR #367, opened 2019-01-30
I2559

EmmaLRussell:


app/docker-compose.yml
```diff
@@ -23,5 +23,11 @@ services:
   db_annex:
     image: docker.montagu.dide.ic.ac.uk:5000/montagu-db:${MONTAGU_DB_VERSION}
     command: /etc/montagu/postgresql.test.conf
+  orderly:
```
EmmaLRussell: Had to add orderly to the development apis since it does the publish step

app/src/main/report/components/Sidebar/PublishSwitch.tsx
```diff
@@ -33,30 +58,57 @@ export class PublishSwitchComponent extends React.Component<Props, undefined> {
...
                     onstyle={"published"}
                     active={this.props.published}/>
-                <UncontrolledTooltip target={"publish"}>
-                    Publish functionality coming soon</UncontrolledTooltip>
             </div>
     }
 }
 
-const mapStateToProps = (state: ReportAppState, props: PublicProps): Partial<Props> => {
+export class PublishSwitchModal extends React.Component<PublishSwitchModalProps, undefined> {
```
EmmaLRussell: I initially tried to make the modal using the bootstrap modal support via reactstrap, but found that Modal required react v16. Attempting to upgrade using npm seemed to break all sorts of things so I've gone with simple handrolled solution for now.

PR #368, opened 2019-02-01
I2443

EmmaLRussell:


app/src/main/report/components/Sidebar/Sidebar.tsx
```diff
@@ -45,19 +45,6 @@ export const mapStateToProps = (state: ReportAppState): Partial<SidebarProps> =>
 
 export const SidebarComponent: React.StatelessComponent<SidebarProps> = (props: SidebarProps) => {
 
-    const changelogLink = props.isReviewer
```
EmmaLRussell: At this point, Sidebar doesn't need isReviewer on its props any more since the Changelog tab is now enabled for all users - but I'm going to leave it on as it will be useful for the Run functionality we'll be adding shortly. (Or should we be looking for isRunner perm for that rather than reviewer?..)

app/src/main/report/components/Reports/ReportChangelog.tsx
```diff
@@ -42,7 +43,7 @@ export const ReportChangelogComponent = (props: ReportChangelogProps) => {
                 <thead className="changelog-header">
                 <tr>
                     <th className="datestring-column">Date</th>
-                    <th>Label</th>
+                    { props.isReviewer && <th>Label</th> }
                     <th>Text</th>
```
littledice: Shall we change "text" to "changes"?

app/config/reporting_api_version
```diff
@@ -1 +1 @@
-master
+i2385
```
EmmaLRussell: Need to change this back to master after i2385 has been merged (montagu API allowing access to changelogs for readers)

PR #369, opened 2019-02-01
I2573

EmmaLRussell:


app/src/main/report/components/Reports/ReportChangelog.tsx
```diff
@@ -22,11 +22,17 @@ export interface ReportChangelogProps extends ReportChangelogPublicProps {
     fetchChangelog: (props: ReportChangelogPublicProps) => void;
 }
 
-export const ReportChangelogComponent = (props: ReportChangelogProps) => {
+export class ReportChangelogComponent extends React.Component<ReportChangelogProps> {
+
+    render() {
+
+        if (this.props.versionChangelog == null) {
```
EmmaLRussell: Moved this from the recompose branch where it was previously, as this meant that componentDidUpdate was not called when it has branched to render LoadingElement instead. Not sure if this is the best approach though!

EmmaLRussell:
> Can we write a test that reproduces this bug/demonstrates that it is fixed?
> Also I think that just adding
> 
> ```
> onDidUpdate: (props: ReportChangelogProps) => {
>             props.fetchChangelog(props)
> }
> ```
> 
> to the recompose logic would fix it, without any of the other changes.

Yes, that's done it, but I had to add onDidUpdate to the d.ts file.

PR #371, opened 2019-02-11
i2602 Changelogs - use a slightly shorter date format and make date a link

littledice:


littledice:
The longer form date in the dropdown doesn't look crowded to me, so I wouldn't change it unless a user requested.

PR #373, opened 2019-02-13
I2560

EmmaLRussell:


app/src/main/shared/services/AbstractLocalService.ts
```diff
@@ -208,8 +208,10 @@ export abstract class AbstractLocalService {
     }
 
     notifyOnErrors(error: any) {
-        this.initOptions();
-        notificationActionCreators.notify(error.toString(), "error")(this.dispatch, null);
+        if (this.options.notificationOnError) {
```
EmmaLRussell: I made this change so that we wouldn't see pop up errors every five seconds when polling for run status if there was a problem contacting the API. But I'm a bit confused by the previous behaviour of the base class. I would have thought that setting notificationOnError in the options would prevent notification, not sure why it was calling initOptions before?

app/src/main/report/reducers/reportsReducer.ts
```diff
@@ -62,6 +65,30 @@ export const reportsReducer = (state = reportsInitialState, action: ReportsActio
...
+                key: action.data.key,
+                status: RunningReportStatusValues.RUNNING_REPORT_STATUS_STARTED,
+                version: null,
+                output: null
+            });
+            return {...state, runningReports: runningReports};
+        case ReportTypeKeys.REPORT_RUN_STATUS_FETCHED:
+            runningReports = state.runningReports;
+            const runningReport = runningReports.find(r => r.key == action.data.key);
+            //Update the object in the state
```
littledice: We shouldn't actually update state objects, only replace: https://redux.js.org/recipes/structuring-reducers/immutable-update-patterns

app/src/main/report/services/ReportsService.ts
```diff
@@ -8,8 +8,7 @@ export class ReportsService extends AbstractLocalService {
     }
 
     getReportVersions(reportId: string) {
-        return this.setOptions({cacheKey: ReportsCacheKeysEnum.versions})
```
littledice: Hmm, maybe `ignoreCache` should be an argument that the method takes, so that we can still use the cache in other contexts?

app/src/main/shared/services/AbstractLocalService.ts
littledice: Not sure either tbh. We should probably put in some work on the error handling at some point, it could do with a refactor and redesign, both the code and the user experience are a bit confusing at the moment!

app/src/main/report/services/ReportsService.ts
EmmaLRussell: The trouble is that the only other place where we call getReportVersions (on page load) we need to ignore the cache too to ensure we pick up freshly run versions (I tried putting in this param for the refresh from RunReport component only, and found my new version disappeared when I navigated to it!). I suppose we could put in some logic to say you can safely ignore the cache if user does not have run perms, or there have been no running reports. Or is the issue that when we set options to not use the cache for fetching, we might still want to update the cache with what we fetch?

littledice: Ah ok, leave this then. It's not a lot of data to request each time.

PR #374, opened 2019-02-15
upgrade typescript and fix errors

littledice:


app/src/test/contrib/components/Responsibilities/BurdenEstimates/DiagnosticSectionTests.tsx
```diff
@@ -56,6 +57,9 @@ describe("Diagnostic section", () => {
 
     beforeEach(() => {
         store = createMockContribStore(testState);
+        sandbox.setStubFunc(EstimatesService.prototype, "getEstimates", () => {
```
littledice: Also just noticed that these tests were trying to `fetch` and throwing tons of warnings because the fetch method wasn't mocked.

PR #378, opened 2019-02-25
add proptypes back in

hillalex:
Honestly don't really get what this `PropTypes` thing is all about or why leaving out this declaration causes the router not to be attached to the context but this has fixed the routing :woman_shrugging:


EmmaLRussell:
Weird! Don't suppose there's any way to write a test for this to check it's fixed as part of the test suite? Were some InternalLinks not getting rendered?

hillalex:
Unfortunately not, it's really something that could only have been caught by manual testing. The whole reason for the `if` clause (https://github.com/vimc/montagu-webapps/pull/378/files#diff-7dbce0b1904e2d6797e52b72858bfbbfL18) is that the `Router` won't be present in a unit text context! The problem was the `Router` object wasn't present so it was rendering a normal `a` tag instead of a router link...

PR #382, opened 2019-03-19
I2654 new upload form

hillalex:
This PR introduces the service, action and reducer methods needed for the new upload form. The form component that uses these will come in a further PR.


app/src/integrationTests/ContribIntegrationTests.tsx
```diff
@@ -5,7 +5,7 @@ import {createMemoryHistory} from 'history';
 
 import {
     CoverageSet, CreateBurdenEstimateSet, DemographicDataset,
-    Disease,
+    Disease, ErrorInfo,
```
EmmaLRussell: I don't think you need this ErrorInfo?

EmmaLRussell: I don't think you need this ErrorInfo?

PR #383, opened 2019-03-20
i2772: Use the smaller container

richfitz:
As discussed yesterday, the webapps should use orderly.server not montagu-orderly for testing

The image is much smaller and built more often


richfitz:
(This was extracted from the changes on the `i2769` branch which do not want merging)

PR #385, opened 2019-03-25
VIMC-2796 move success and error alerts to top level component

hillalex:
As I was testing on UAT it came to my attention that bc of the weird 2 step form the error/success alerts after uploading a set of estimates were in the wrong place. 

They should live one component level up from the `PopulateBurdenEstimatesForm`, in the component that renders both that and the `CreateBurdenEstimateSetForm`. This component was still expecting error/success messages to come from a query string, so have changed it over to look in the store state instead.

### DIFF
--- a/app/src/main/contrib/components/Responsibilities/BurdenEstimates/UploadBurdenEstimatesForm.tsx
+++ b/app/src/main/contrib/components/Responsibilities/BurdenEstimates/UploadBurdenEstimatesForm.tsx
@@ -1,11 +1,14 @@
 import * as React from "react";
-import {helpers} from "../../../../shared/Helpers";
 import {Alert} from "reactstrap";
-import {BurdenEstimateSet, ErrorInfo, Result} from "../../../../shared/models/Generated";
+import {BurdenEstimateSet, ErrorInfo} from "../../../../shared/models/Generated";
 import {CreateBurdenEstimateSetForm} from "./CreateBurdenEstimateSetForm";
 import {PopulateEstimatesForm} from "./PopulateBurdenEstimatesForm";
+import {Dispatch} from "redux";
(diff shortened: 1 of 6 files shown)


app/src/main/contrib/components/Responsibilities/BurdenEstimates/UploadBurdenEstimatesForm.tsx
```diff
@@ -14,48 +17,32 @@ export interface UploadBurdenEstimatesFormComponentProps {
...
+    errors: ErrorInfo[];
+    resetPopulateState: () => void;
 }
 
-export class UploadBurdenEstimatesForm extends React.Component<UploadBurdenEstimatesFormComponentProps, State> {
-    private uploadSuccessMessage = "Success! You have uploaded a new set of burden estimates";
-
-    constructor(props: UploadBurdenEstimatesFormComponentProps) {
-        super(props);
-        const result = helpers.ingestQueryStringAndReturnResult();
```
hillalex: This is a vestige from the redirect flow where the user was redirected to this page with error/success messages encoded in a query string. Success/error results are now populated my mapping the global state to props below.

app/src/test/contrib/components/Responsibilities/BurdenEstimates/PopulateBurdenEstimateFormComponentTests.tsx
```diff
@@ -213,15 +213,6 @@ describe("Populate Burden Estimates Form Component tests", () => {
         expect(newState.file.fileName).to.eq("test.csv")
     });
 
-    it("shows success message on file population success", () => {
```
hillalex: This component is actually immediately hidden on success, so the success message should not be shown here.

PR #386, opened 2019-04-03
reintroduce sorting with test

hillalex:
Fixes regression with report versions sorted in the wrong order.


app/src/main/report/components/ReportsList/ReportListTable.tsx
```diff
@@ -182,7 +182,7 @@ export class ReportsListTable extends React.Component<ReportsListTableProps, any
             </ButtonGroup>
             <ReactTable
                 pivotBy={["name"]}
-                defaultSorted={[{id: "version", desc: true}]}
+                defaultSorted={[{id: "version", desc: false}]}
```
EmmaLRussell: Wait..what? We want the sort to be descending don't we, so the latest goes first? That's what the (passing!) test is testing for. I'm sure it works, I'm just not sure how. Maybe a comment would be useful?

PR #388, opened 2019-05-02
VIMC-2865 Remove use of local storage for storing tokens

EmmaLRussell:
Remove use of local storage for storing tokens and replace it with getting authenticated user information from api


app/src/main/shared/actions/authActionCreators.ts
```diff
@@ -33,25 +34,39 @@ export const authActionCreators = {
...
-            const token = localStorageHandler.get("accessToken");
-            if (token) {
-                const inflated = jwtTokenAuth.inflateToken(token);
-                const decoded = inflated && jwtTokenAuth.decodeToken(inflated);
+    loadAuthenticatedUser() {
+        // fetch details of currently logged in user from API.
+        return async (dispatch: Dispatch<any>, getState: () => GlobalState) => {
+            try {
+                const userResponse = await(new AuthService(dispatch, getState)).getCurrentUser();
+                //TODO: since we should always have the modelling groups for the current user, either from the API here
```
EmmaLRussell: Does this sound right to you? I remember adding this endpoint to get modelling groups for the current user, but now not sure why we needed it if the modelling groups were already embedded in the token! Arguably we could roll this endpoint into the currentUser one if this is the only place it's needed. Shall I create another ticket for that?

app/src/main/shared/modules/jwtTokenAuth.ts
```diff
@@ -58,18 +58,17 @@ export const jwtTokenAuth = {
...
-        return {
-            receivedBearerToken: true,
-            receivedCookies: false,
-            bearerToken: token,
-            isAccountActive: permissions.some((x: string) => x == "*/can-login"),
-            isModeller: modellingGroups.length > 0,
-            username: decoded.sub,
+
+        const result = loadAuthState(
+            decoded.sub, //username
```
hillalex: I think the commenting of args here suggests it'd be better to use an object as args for this function, i.e.
```
interface AuthStateOptions {
    username: string,
    receivedBearerToken: boolean,
    receivedCookies: boolean,
    bearerToken: string,
    permissions: string[],
    modellingGroups: string[]
}

export function loadAuthState(options: AuthStateOptions)
```

Advantages to doing it this way - can't make mistakes with param order, self-documenting, and better intellisense :smile:

PR #389, opened 2019-05-13
VIMC-2815 deprecate portal logins

hillalex:
Contains commits from https://github.com/vimc/montagu-webapps/pull/390 so merge that one first

* remove login pages and tests
* remove password related pages and tests
* redirect users to Montagu login page after a failed login

I also simplified the `AuthState` so that instead of `hasBearerToken` and `hasCookies` there is just a single boolean `loggedIn` which is only true after cookies are set.


hillalex:
Sure! https://github.com/vimc/montagu-webapps/pull/389/commits/333b23b0c1ba897e56a84f6a9aeeb5d95df70034

hillalex:
Ok, this was a bit of an annoying one. A race condition meant that even though the call to generate the login page breadcrumbs was being fired first, it was completing after the action to generate the main menu breadcrumbs.

Contrib was working fine because the homepage was loading data, meaning the breadcrumb generation was happening slightly later. 

To fix it for admin I've added a timeout to force the admin page to load fractionally more slowly: https://github.com/vimc/montagu-webapps/pull/389/commits/8b2cae5d57408c640b9b8006f51b966b3dbe5d8a#diff-4d1db3532bbd7572f13c8c6e3dd35f50R23

To fix it for reports I've changed things over so that the homepage makes the call to get reports, rather than the child component. Also realised that we haven't yet migrated the reportle over to using the page pattern we have in the other two portals so I migrated the homepage to use the same pattern (although of course pointless to migrate the other pages since we'll soon be replacing the lot!)

https://github.com/vimc/montagu-webapps/pull/389/commits/8b2cae5d57408c640b9b8006f51b966b3dbe5d8a

PR #397, opened 2019-08-06
VIMC-3020 Add missing model metadata columns

EmmaLRussell:
This isn't quite everything in the ticket, since we don't yet have the floating pop-outs for scenario and country lists (or outcome explanations) - unless those should go in a separate ticket, but I thought it was about time you had a look at it.


app/src/main/admin/components/ModellingGroups/Models/ModelMetaTable.tsx
```diff
@@ -102,20 +102,20 @@ export class ModelMetaTableComponent extends React.Component<ModelMetaProps, Sta
...
-                    {this.createHeader("code", "Code")}
-                    {this.createHeader("gender", "Gender")}
-
-                    {this.createHeader("max_countries", "Max Countries")}
-                    {this.createHeader("years", "Years")}
-                    {this.createHeader("ages", "Ages")}
-                    {this.createHeader("cohorts", "Cohorts")}
-                    {this.createHeader("outcomes", "Outcomes", true)}
-                    {this.createHeader("has_dalys", "DALYs")}
+                    {this.createHeader("modelling_group", "Group", "7em")}
```
EmmaLRussell: Maybe there's a cleverer way to do this but I couldn't think of one! This forces the table to scroll at narrow widths

app/scripts/clear-docker.sh
```diff
@@ -1,3 +1,5 @@
 docker stop $(docker ps -aq)
 docker rm $(docker ps -aq)
 docker network prune --force
+
+docker ps
```
EmmaLRussell: Hm, I put this in to try and diagnose a bit more about the weird db containers that still seem to be getting in the way on some builds. This script *should* be stopping running containers. But then the next time it ran fine.....

richfitz: try `docker rm -f`

PR #399, opened 2019-08-07
VIMC-1222 combine create and populate steps into 1 form

hillalex:
Have combined the two step burden estimate upload into one without changing the API spec. On form submit it creates the set, then requests an upload token, then uploads the file of estimates. It does mean the upload form is even more horrendous than it was though (forms - so simple in html/vanilla js, so damn complicated in React :roll_eyes: )


app/src/main/contrib/components/Responsibilities/BurdenEstimates/UploadBurdenEstimatesForm.tsx
```diff
@@ -1,86 +1,345 @@
...
+const FileUploadClient = require('resumablejs');
 
-export interface UploadBurdenEstimatesFormPublicProps {
+export interface UploadEstimatesState {
+    file: ResumableFile,
+    progress: number,
+    isUploading: boolean,
+    uploadErrors: ErrorInfo[],
+    fileValidationResult: CustomValidationResult,
+    metadata: BurdenEstimateSetType
```
hillalex: The contents of this component are mostly just the old `PopulateEstimateForm` renamed, but unfortunately git can't give a nice diff because it's also been renamed, so I'll try to point out the changes.

Here all that's changed is the `metadata` field has been added and the `validationResult` renamed to `fileValidationResult`.

app/src/main/contrib/components/Responsibilities/BurdenEstimates/UploadBurdenEstimatesForm.tsx
```diff
@@ -1,86 +1,345 @@
...
-export interface UploadBurdenEstimatesFormComponentProps extends UploadBurdenEstimatesFormPublicProps {
-    hasSuccess: boolean;
-    errors: ErrorInfo[];
+export interface UploadEstimatesProps extends UploadEstimatesPublicProps {
+    createUploadClient: (settings: ConfigurationHash) => Resumable,
+    populateEstimateSet: (token: string) => void,
     resetPopulateState: () => void;
+    url: string;
+    uploadToken: string;
+    createBurdenEstimateSet: (data: BurdenEstimateSetType) => void;
```
hillalex: The change here is that `requestUploadToken` has become `createBurdenEstimateSet` (which itself creates the set and then gets the upload token)

app/src/main/contrib/components/Responsibilities/BurdenEstimates/UploadBurdenEstimatesForm.tsx
```diff
@@ -1,86 +1,345 @@
...
+            });
+        });
+    };
+
+    startUpload = () => {
+
+        this.props.resetPopulateState();
+        this.props.createBurdenEstimateSet(this.state.metadata);
+
+        const tryUpload = setInterval(() => {
```
hillalex: This method now first creates the burden estimate set (which will populate the setId and upload token in the store state) and then waits for those props to exist before starting the file upload. It's slightly ridiculous because really this is a synchronous process that is being forced into an async one by redux, but it seemed like the more minimal refactor. Alternatively we could refactor this component to manage all it's own state and avoid redux altogether...

app/src/main/contrib/components/Responsibilities/BurdenEstimates/UploadBurdenEstimatesForm.tsx
```diff
@@ -1,86 +1,345 @@
...
+
+    render() {
+
+        return (
+            <div>
+                <div className="progress" style={{display: this.state.progress === 0 ? "none" : "block"}}>
+                    <div className={`progress-bar ${this.state.uploadErrors.length > 0 ? "bg-danger" : "bg-success"}`}
+                         style={{width: this.state.progress + '%'}}>&nbsp;</div>
+                </div>
+                <div className="row">
```
hillalex: Added these 2 new fields

app/src/main/contrib/components/Responsibilities/BurdenEstimates/UploadBurdenEstimatesForm.tsx
```diff
@@ -1,86 +1,345 @@
...
-            <Alert color="success" isOpen={this.props.hasSuccess} toggle={this.props.resetPopulateState}>
-                Success! You have uploaded a new set of burden estimates
-            </Alert>
-
-            {createForm}
-            {this.renderUploadForm()}
-        </div>;
+    private uploadClient: Resumable;
+    private uploaderElement: Element;
+    private options = [
```
hillalex: Options for the dropdown added

app/src/main/contrib/components/Responsibilities/BurdenEstimates/UploadBurdenEstimatesForm.tsx
```diff
@@ -1,86 +1,345 @@
...
+
+    };
+
+    cancelUpload = () => {
+        this.resetUploadState(true);
+        this.uploadClient.cancel();
+    };
+
+    uploadDisabled = () => {
+        return this.submitDisabled()
```
hillalex: We allow submission when all fields are filled in, but until the set has been created the url will be null (because the newly created set id forms part of the url) so we need to wait for the url to be non null before beginning file upload

app/src/main/contrib/components/Responsibilities/BurdenEstimates/UploadBurdenEstimatesForm.tsx
```diff
@@ -1,86 +1,345 @@
...
+
+    submitDisabled = () => {
+        return this.state.isUploading
+            || this.state.file == null
+            || (this.state.fileValidationResult && !this.state.fileValidationResult.isValid)
+            || !this.state.metadata.details
+            || !this.state.metadata.type
+    };
+
+    onDetailsChange = (e: FormEvent<HTMLInputElement>) => {
```
hillalex: This and onTypeChange added to support the new fields

app/src/main/contrib/components/Responsibilities/BurdenEstimates/UploadBurdenEstimatesForm.tsx
```diff
@@ -1,86 +1,345 @@
...
+    };
+
+export const mapStateToProps: (state: ContribAppState, props: UploadEstimatesPublicProps) => Partial<UploadEstimatesProps>
+    = (state: ContribAppState, props: UploadEstimatesPublicProps) => {
     return {
         ...props,
-        errors: state.estimates.populateErrors,
-        hasSuccess: state.estimates.hasPopulateSuccess
+        url: state.estimates.uploadToken == null || state.estimates.populatingSetId == null ? null :
+            `${settings.apiUrl()}/modelling-groups/${props.groupId}/responsibilities/${props.touchstoneId}/${props.scenarioId}/estimate-sets/${state.estimates.populatingSetId}/actions/upload/${state.estimates.uploadToken}/`,
```
hillalex: The setId is now coming from store state (see https://github.com/vimc/montagu-webapps/pull/399/files#diff-fdc00e5b43ae62249845b96dcc1d4b33R86)

hillalex:
Just to note, the 'You registered how you calculated your central estimates' message shows up when the `responsibility.current_estimate_set` is an empty set. After this form change is finished we should stop this from ever happening - ticket for this here https://vimc.myjetbrains.com/youtrack/issue/VIMC-1564

hillalex:
The cancel button is if you want to cancel a long-running upload. You *should* see it enabled while uploading a large file!

app/src/main/contrib/components/Responsibilities/BurdenEstimates/UploadBurdenEstimatesForm.tsx
EmmaLRussell: Do you need to do this with setInterval? Could mapStateToProps check if the burden estimate set has just been created and is ready for data (if there's a populatingSetId in the state?), and componentDidUpdate initiate the upload there, clearing populatingSetId immediately after?

hillalex: Omg yes you're right, that would be basically the same logic. Thank you!

app/src/main/contrib/components/Responsibilities/BurdenEstimates/UploadBurdenEstimatesForm.tsx
```diff
@@ -340,6 +332,5 @@ export const mapDispatchToProps:
 
 export const UploadEstimatesForm =
     compose<UploadEstimatesProps, UploadEstimatesPublicProps>(
-        connect(mapStateToProps, mapDispatchToProps),
-        branch((props: UploadEstimatesProps) => props.populatingInProgress, renderComponent(LoadingElement))
```
hillalex: This was causing the component to unmount/remount when `populatingInProgress` changed, hence why the fields were getting cleared. Have moved the loading spinner to just replace the buttons while populating is in progress, rather than replacing the whole component.)

app/src/main/contrib/reducers/estimatesReducer.ts
```diff
@@ -66,7 +66,8 @@ export const estimatesReducer = (state = estimatesInitialState, action: Estimate
                 ...state,
                 hasPopulateSuccess: false,
                 populateErrors: [],
-                populatingSetId: null
+                populatingSetId: null,
+                uploadToken: null
```
hillalex: The token not being re-set was causing the bug you saw with the "has this token been used already" message

PR #402, opened 2019-08-12
update test to be more flexible

hillalex:
Also fixed the alignment of buttons on the responsibilities overview page cos it looked weird :sweat_smile:


app/src/integrationTests/ContribIntegrationTests.tsx
```diff
@@ -251,7 +251,7 @@ class ContributionPortalIntegrationTests extends IntegrationTestSuite {
             const result = await (new EstimatesService(this.store.dispatch, this.store.getState))
                 .createBurden(groupId, touchstoneVersionId, scenarioId, data);
 
-            expect(result).to.eql("http://api:8080/v1/modelling-groups/test-group/responsibilities/test-1/yf-1/estimate-sets/1/");
```
hillalex: This exact url varies depending on where the test is run! Hence http://teamcity.montagu.dide.ic.ac.uk:8111/viewLog.html?buildId=57193&tab=buildResultsDiv&buildTypeId=montagu_deploy

PR #403, opened 2019-08-12
add filename to current estimate set summary

hillalex:
Should only be merged after https://github.com/vimc/montagu-api/pull/403

### DIFF
--- a/app/config/api_version
+++ b/app/config/api_version
@@ -1,1 +1,1 @@
-VIMC-3062
+master
--- a/app/config/api_version
+++ b/app/config/api_version
@@ -1,1 +1,1 @@
-master
+VIMC-3062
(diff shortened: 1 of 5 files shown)


app/config/api_version
```diff
@@ -1 +1 @@
-master
+VIMC-3062
```
hillalex: Once https://github.com/vimc/montagu-api/pull/403 merged this should be returned to master

app/src/main/contrib/components/Responsibilities/Overview/List/CurrentEstimateSetSummary.tsx
```diff
@@ -58,10 +58,10 @@ export class CurrentEstimateSetSummary extends React.Component<CurrentEstimateSe
         return <div className={`mt-3 alert alert-${alertType} text-left`}>
             {estimateText}
             {url && [
-                <FileDownloadButton href={url} className={"float-right"}>
+                <FileDownloadButton href={url} className={"float-right"} key={`download-${estimateSet.id}`}>
```
hillalex: I noticed React was complaining that you need to use a key when rendering an array of elements as here

PR #405, opened 2019-08-14
Vimc 3085 Show current and obsolete models in separate tables

EmmaLRussell:
Contains changes from https://github.com/vimc/montagu-webapps/pull/404 which should be merged first

### DIFF
--- a/app/src/main/admin/components/ModellingGroups/Models/ModelMetaTable.tsx
+++ b/app/src/main/admin/components/ModellingGroups/Models/ModelMetaTable.tsx
@@ -116,7 +116,7 @@ export class ModelMetaTableComponent extends React.Component<ModelMetaProps, Sta
     createExpectationHeaders = () => {
         return <React.Fragment>
             {this.createHeader("scenario_count", "Scenarios", "8.5em")}
-            {this.createHeader("years", "Years", "6.5em")}}
+            {this.createHeader("years", "Years", "6.5em")}
             {this.createHeader("ages", "Ages", "6em")}
             {this.createHeader("cohorts", "Cohorts", "7.5em")}
             {this.createHeader("outcomes", "Outcomes", "8.5em")}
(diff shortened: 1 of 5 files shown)


app/src/main/admin/components/ModellingGroups/Models/ModelMetaTable.tsx
```diff
@@ -110,82 +115,88 @@ export class ModelMetaTableComponent extends React.Component<ModelMetaProps, Sta
...
     render() {
 
+        const obsolete = this.props.obsoleteModels;
+
+        if (obsolete && this.state.data.length == 0) {
+            return "";
+        }
+
         return <div>
+            {!obsolete || <p>The following obsolete models were also found.</p>}
```
hillalex: Is there a reason to use inverted logic here? If not I think `{obsolete && <p></p>}` is easier to read.

app/src/main/admin/components/ModellingGroups/Models/ModelMetaTable.tsx
```diff
@@ -110,82 +115,88 @@ export class ModelMetaTableComponent extends React.Component<ModelMetaProps, Sta
...
+                                </td>}
                             <td data-title="code">{model.code}</td>
                             <td data-title="gender">{model.gender ? model.gender : "NA"}</td>
                             <td data-title="max_countries">{model.max_countries}
                                 {countriesDetailsLink}</td>
-                            <td data-title="years">{model.years}</td>
-                            <td data-title="ages">{model.ages}</td>
-                            <td data-title="cohorts">{model.cohorts}</td>
-                            <td data-title="outcomes">{model.outcomes}
+                            {obsolete || <td data-title="years">{model.years}</td>}
```
hillalex: Maybe the creation of these 5 columns could be moved into their own method so this is a little easier to parse?
So you end up with something like `{obsolete} || expecationsColumns()` instead of repeating the conditional. Similarly for the corresponding headers.

app/src/main/admin/components/ModellingGroups/Models/ModelMetaTable.tsx
```diff
@@ -200,12 +211,24 @@ export class ModelMetaTableComponent extends React.Component<ModelMetaProps, Sta
...
-    const expectationNotFound  = "Error: expectation not found for model";
+    const includedModels = state.groups.models.filter((model) => {
+        //A model is considered obsolete if it has no expectations
+        return (props.obsoleteModels && modelExpectations[model.id].length == 0) ||
+            (!props.obsoleteModels && modelExpectations[model.id].length > 0)
+    });
 
     return {
-        models: state.groups.models.map(m => {
+        obsoleteModels: props.obsoleteModels,
```
hillalex: This keeps throwing me off as I expect `obsoleteModels` to be an array of obsolete models! Could we make it something more obviously a bool like `obsolete` or `includeObsoleteOnly` or something?

PR #406, opened 2019-08-15
don't show filename if it doesn't exist

hillalex:
We don't have the filename stored for old estimate sets. We shouldn't show the filename messages when the filename is null.


hillalex:
2 failing tests related to the change in the expecations endpoint which I think are fixed in https://github.com/vimc/montagu-webapps/pull/404

PR #408, opened 2019-08-19
fix issues with form upload

hillalex:
This fixes the following bug:
1. upload a file through the burden estimate upload form
2. change the metadata (or do nothing) and hit upload again without changing the file
3. upload button gets greyed out but nothing happens

### DIFF
--- a/app/src/main/contrib/actions/estimatesActionCreators.ts
+++ b/app/src/main/contrib/actions/estimatesActionCreators.ts
@@ -34,7 +34,6 @@ export const estimatesActionCreators = {
             if (newSetUrl) {
                 const setId = newSetUrl.split("/").filter(Number).pop();
                 dispatch(estimatesActionCreators.getUploadToken(setId));
-                dispatch(responsibilitiesActionCreators.refreshResponsibilities());
                 dispatch({
                     type: EstimateTypes.SET_CREATED,
                     data: setId
@@ -64,10 +63,11 @@ export const estimatesActionCreators = {
                 data: true
             } as PopulatingEstimateSet);
 
-            const ids = mapStateToPropsHelper.getResponsibilityIds(getState());
+            const state = getState();
+            const ids = mapStateToPropsHelper.getResponsibilityIds(state);
 
             const result = await (new EstimatesService(dispatch, getState))
-                .populateEstimatesFromFile(ids.groupId, ids.touchstoneId, ids.scenarioId, ids.estimateSetId, uploadToken);
+                .populateEstimatesFromFile(ids.groupId, ids.touchstoneId, ids.scenarioId, state.estimates.populatingSetId, uploadToken);
 
             dispatch(this._estimateSetPopulated(result));
             dispatch(responsibilitiesActionCreators.refreshResponsibilities())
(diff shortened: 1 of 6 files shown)


app/src/main/contrib/actions/estimatesActionCreators.ts
```diff
@@ -64,10 +63,11 @@ export const estimatesActionCreators = {
                 data: true
             } as PopulatingEstimateSet);
 
-            const ids = mapStateToPropsHelper.getResponsibilityIds(getState());
+            const state = getState();
+            const ids = mapStateToPropsHelper.getResponsibilityIds(state);
 
             const result = await (new EstimatesService(dispatch, getState))
-                .populateEstimatesFromFile(ids.groupId, ids.touchstoneId, ids.scenarioId, ids.estimateSetId, uploadToken);
+                .populateEstimatesFromFile(ids.groupId, ids.touchstoneId, ids.scenarioId, state.estimates.populatingSetId, uploadToken);
```
hillalex: This wasn't actually affecting behaviour because this endpoint doesn't validate the set id by the looks of things, it just uses the upload token to identify the intended set. We should probably change this!

app/src/main/contrib/actions/estimatesActionCreators.ts
```diff
@@ -34,7 +34,6 @@ export const estimatesActionCreators = {
             if (newSetUrl) {
                 const setId = newSetUrl.split("/").filter(Number).pop();
                 dispatch(estimatesActionCreators.getUploadToken(setId));
-                dispatch(responsibilitiesActionCreators.refreshResponsibilities());
```
hillalex: Just not needed at this stage

app/src/main/contrib/components/Responsibilities/BurdenEstimates/UploadBurdenEstimatesForm.tsx
```diff
@@ -141,7 +141,6 @@ export class UploadEstimatesFormComponent extends React.Component<UploadEstimate
 
         this.uploadClient.on('fileSuccess', () => {
             this.resetUploadState(false);
-            this.props.createBurdenEstimateSet(this.state.metadata);
```
hillalex: This has already been called! Was creating an extra set here

app/src/main/contrib/components/Responsibilities/BurdenEstimates/UploadBurdenEstimatesForm.tsx
```diff
@@ -157,14 +156,20 @@ export class UploadEstimatesFormComponent extends React.Component<UploadEstimate
...
             });
         });
     };
 
     componentDidUpdate(prevProps: Readonly<UploadEstimatesProps>) {
-        if (prevProps.url == null && this.props.url != null) {
-            this.uploadClient.upload();
+        if (prevProps.url != this.props.url && this.props.url != null) {
+            if (this.uploadClient.files[0].progress(false) == 1){
+                this.uploadClient.files[0].retry();
```
hillalex: This was the cause of the weird behaviour I was seeing. If the file has already been uploaded, and the user just hits "upload" again without changing the file (e.g. because they've changed the metadata), the client doesn't reupload it because it thinks it's already done that successfully. In reality the url will have changed (the client doesn't seem to detect this and invalidate progress) so it needs to re-upload it to thew new url.

hillalex:
I've deployed this branch onto UAT for testing

PR #414, opened 2019-10-30
VIMC-3210 use pdfs for guidance

hillalex:
Adds pdfs containing touchstone guidance for 2019 touchstones.


app/src/main/contrib/actions/estimatesActionCreators.ts
```diff
@@ -103,7 +103,7 @@ export const estimatesActionCreators = {
 
             return dispatch({
                 type: EstimateTypes.BURDEN_ESTIMATES_FETCHED,
-                data: {burdens: data, type: outcome, setId: setId}
+                data: {burdens: Object.freeze(data), type: outcome, setId: setId}
```
hillalex: Just an opportunistic tweak!

app/src/main/contrib/components/Responsibilities/Guidance/ResponsibilityGuidanceModelInputsPage.tsx
```diff
@@ -9,30 +9,35 @@ import {settings} from "../../../../shared/Settings";
...
         if (settings.is2017Touchstone(this.props.touchstoneVersion.id)) {
-            return <ResponsibilityGuidanceModelInputsContent2017 />
-        }
-        else {
-            return <ResponsibilityGuidanceModelInputsContentLatest />
+            return <ResponsibilityGuidanceModelInputsContent2017/>
+        } else if (settings.is2019Touchstone(this.props.touchstoneVersion.id)) {
+            const inputsPdf = require("./content/guidance-2019-inputs.pdf");
+            return <PageArticle title={"Guidance on model inputs: coverage and demographic data"}>
+                <a href={inputsPdf} target={"_blank"}>
```
hillalex: This page won't actually be linked to for 2019 touchstones, but have added this logic just in case someone visits the url directly (unlikely).

hillalex:
> Looks good! I suppose if this is how we're going to want to do things in all future touchstones, maybe the longterm approach would be to have a naming convention for the expected pdf by touchstone year, and rather than making a special case for 2019, make a special case for in the inline HTML for 2018 like we have for 2017 and default to the pdf link. But maybe that's something to do in 2020!

Yeah agreed, I was thinking exactly the same. But I figure next time we touch the code!

PR #416, opened 2019-11-07
vimc-3222 Removed unnecessary guidance links.

EmmaLRussell:


app/src/integrationTests/AdminIntegrationTests.tsx
```diff
@@ -297,8 +297,8 @@ class AdminIntegrationTests extends IntegrationTestSuite {
             expect(response.status).to.equal(200);
             const result = await response.text();
 
-            expect(result).to.eq("scenario,set_name,vaccine,gavi_support,activity_type,country_code,country,year,age_first,age_last,age_range_verbatim,target,coverage\n"
-                + "yf-1,Test set,yf,no vaccine,none,ATL,Atlantis,1970,1,2,1-2,1000,1000\n")
+            expect(result).to.eq("scenario,set_name,vaccine,gavi_support,activity_type,country_code,country,year,age_first,age_last,age_range_verbatim,target,coverage,gender\n"
```
EmmaLRussell: These integration tests were failing because of the changes for gender in coverage data in a previous PR. So I guess webapp build/test isn't triggered by API build and counted as an API check..

hillalex: Ah yes, it is auto triggerred in TC (http://teamcity.montagu.dide.ic.ac.uk:8111/viewLog.html?buildId=67689&tab=buildResultsDiv&buildTypeId=montagu_webapps_build_contrib) but not reported as a check in GitHub - not sure if there's a way to add that? Would be good if so.

PR #417, opened 2019-11-21
VIMC-3254 hide wide format

hillalex:
Wide format has a bug for campaign data, so we're hiding the format control for all diseases that have campaigns in the latest touchstone until this is fixed.

### DIFF
--- a/app/src/main/shared/settings/default.ts
+++ b/app/src/main/shared/settings/default.ts
@@ -14,7 +14,9 @@ export const settings: Settings = {
         return !this.nonStochasticTouchstones.some((ts: string) => touchstoneId.indexOf(ts) === 0);
     },
     hideWideFormat: function (disease) {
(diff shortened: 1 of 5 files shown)


app/src/main/shared/settings/default.ts
```diff
@@ -13,6 +13,9 @@ export const settings: Settings = {
         //is touchstone id NOT a version of any name in array of non-stochastic touchstones
         return !this.nonStochasticTouchstones.some((ts: string) => touchstoneId.indexOf(ts) === 0);
     },
+    hideWideFormat: function (disease) {
+        return ["MenA", "JE", "YF", "HPV", "measles", "rubella"].indexOf(disease) > -1
```
EmmaLRussell: I think Rubella and Measles are in the database with first letter upper-cased

hillalex: oh thanks for checking, in fact lets make these case insensitive for robustness

PR #419, opened 2020-01-10
VIMC-3340 update test login mechanism and delete unused code

hillalex:
Although the app code doesn't rely on the JSON webtokens, the integration tests still do! This PR updates them and removes unused auth related code.


app/src/integrationTests/IntegrationTest.ts
```diff
@@ -61,62 +54,23 @@ export abstract class IntegrationTestSuite {
...
+                initialAuthState.bearerToken = response.access_token;
+                await this.store.dispatch(authActionCreators.loadAuthenticatedUser());
             });
 
             this.addTestsToMocha();
         });
     }
 }
 
-export async function firstDownloadPromise(rendered: ReactWrapper) {
```
hillalex: Just noticed a bunch of helpers methods that were used in the reportle tests are now not in use.

PR #421, opened 2020-01-17
VIMC-3348 add code coverage

hillalex:
To do this I've migrated us away from the deprecated mocha-webpack package to jest. Tests are now running a lot faster and the interface for running individual tests is much nicer, see updated README https://github.com/vimc/montagu-webapps/pull/421/files#diff-069aca75954344cbe412c9b31002d366R27

Obviously this PR contains a ridiculous number of changes because of the test migrations, but I've indicated in the file list the point at which the non-test files stop.


app/src/test/ActionCreatorTestHelpers.ts
```diff
@@ -1,7 +1,8 @@
 import {createMockStore} from "./mocks/mockStore";
```
hillalex: From this point onwards in the PR all changes are just migrating from chai/mocha syntax to jest.

hillalex:
Hmm, just noticed that the test count in TeamCity has gone down. Let me investigate

hillalex:
Mystery solved, some test files were badly named and not getting picked up by jest. Test count is back up now and this can be reviewed.

PR #422, opened 2020-01-17
VIMC-3363 remove Sinon

hillalex:
Remove Sinon from the project, replace with Jest.

~~Contains commits from https://github.com/vimc/montagu-webapps/pull/421 so please merge that one first.~~


app/src/main/contrib/components/ChooseAction/ChooseActionPage.tsx
```diff
@@ -12,10 +12,6 @@ export interface ChooseActionPageLocationProps {
 
 export class ChooseActionPageComponent extends React.Component<PageProperties<ChooseActionPageLocationProps>> {
 
-    componentDidMount() {
```
hillalex: Changing the test for this from `expect(mock.called)` to `expect(mock.calls.length).toBe(1)` revealed that this was being called twice! Once in the `PageActionCreators`, and once superfluously here. There are no other changes to app code here, only to tests.

PR #423, opened 2020-01-20
vimc-3230 Run orderlyweb as part of dependencies

EmmaLRussell:


EmmaLRussell:
So this works OK for running up the apps locally and adding a new user - the user is added to the orderly db. However, an integration test is failing on TeamCity and I'm not sure why, as it looks like it's running up the same dependencies as when running manually. Could this be a docker-within-docker issue?

app/scripts/run-dependencies.sh
```diff
@@ -13,6 +13,22 @@ docker run --rm \
...
+rm $PWD/demo -rf
+rm $PWD/git -rf
+mkdir $PWD/demo
+ORDERLY_IMAGE="vimc/orderly:master"
+OW_MIGRATE_IMAGE="vimc/orderlyweb-migrate:master"
+OW_CLI_IMAGE="vimc/orderly-web-user-cli:master"
+
+#create orderly db
+docker pull $ORDERLY_IMAGE
+docker run --rm --entrypoint create_orderly_demo.sh -v "$PWD:/orderly" -u $UID -w /orderly $ORDERLY_IMAGE .
```
hillalex: This is a docker in docker issue - volume mounts here relate to locations on the host machine because the docker sock is mounted from the host. Basically this apporach won't work, I'll have a little think about best alternatives...

hillalex:
Made a couple of changes to get this working:
1. run deps outside of the container (this will work fine without further changes because everything is running on the host network anyway)
2. added ./scripts/stop-dependencies.sh as a final step in TC configuration to make sure deps are torn down
3. added the orderly demo artifact from the main build in TC as a dependency for this build, so no need to recreate the demo data on TC

For points 2 and 3 see the builld steps and depencencies settings on TC: http://teamcity.montagu.dide.ic.ac.uk:8111/admin/editBuild.html?id=buildType:montagu_webapps_build_contrib

EmmaLRussell:
> Made a couple of changes to get this working:
> 
>     1. run deps outside of the container (this will work fine without further changes because everything is running on the host network anyway)
> 
>     2. added ./scripts/stop-dependencies.sh as a final step in TC configuration to make sure deps are torn down
> 
>     3. added the orderly demo artifact from the main build in TC as a dependency for this build, so no need to recreate the demo data on TC
> 
> 
> For points 2 and 3 see the builld steps and depencencies settings on TC: http://teamcity.montagu.dide.ic.ac.uk:8111/admin/editBuild.html?id=buildType:montagu_webapps_build_contrib

Thanks for that. I thought that might be the only solution, but maybe a nuclear option since it's preferable not to run on metal in TC.  Good point on not making orderly demo if not needed. Want me to merge this?

hillalex:
> > Made a couple of changes to get this working:
> > ```
> > 1. run deps outside of the container (this will work fine without further changes because everything is running on the host network anyway)
> > 
> > 2. added ./scripts/stop-dependencies.sh as a final step in TC configuration to make sure deps are torn down
> > 
> > 3. added the orderly demo artifact from the main build in TC as a dependency for this build, so no need to recreate the demo data on TC
> > ```
> > 
> > 
> > For points 2 and 3 see the builld steps and depencencies settings on TC: http://teamcity.montagu.dide.ic.ac.uk:8111/admin/editBuild.html?id=buildType:montagu_webapps_build_contrib
> 
> Thanks for that. I thought that might be the only solution, but maybe a nuclear option since it's preferable not to run on metal in TC. Good point on not making orderly demo if not needed. Want me to merge this?

The deps are still all running in docker, not on metal! Was just waiting to merge this until the API change was in.

PR #432, opened 2020-07-13
vimc-4120 add build to BuildKite

hillalex:
This also removes the TeamCity build scripts; once merged we can remove the TC webhook.

* adds BuildKite directory with scripts and pipeline
* moves docker subdirectory to top level of repo
* moves config to top level of repo
* replaces all references to the local docker registry with the public docker hub org
* removes TeamCity scripts
* replaces Travis build badge with BuildKite build badge


docker/shared-build-env.dockerfile
```diff
@@ -35,25 +35,27 @@ WORKDIR /workspace
 
 # Install Gradle wrapper
 RUN mkdir -p src/webmodels/
-COPY ./src/webmodels/gradlew src/webmodels/
-COPY ./src/webmodels/gradle/ src/webmodels/gradle/
+COPY ./app/src/webmodels/gradlew src/webmodels/
```
hillalex, line 38: These changes are due to the fact that it was convenient to move the `docker` subdirectory up to the top level (previously at `app/docker`)

docker/shared-build-env.dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:8
+FROM node:8-stretch
```
hillalex, line 1: This was simply not working as the node:8 image is based on jessie rather than stretch. Not quite sure why this was working on TeamCity, but possibly it was in the docker cache on all the agents so hadn't been refreshed in a long time.

hillalex:
Ok, build is passing reliably now, think the problem was old dependency containers that I wasn't cleaning up after the build.

docker/shared-build-env.dockerfile
```diff
@@ -0,0 +1,39 @@
+FROM vimc/node-docker-openjdk:master
```
hillalex, line 1: This is mostly just copied from its previous location but github hasn't worked that out

docker/build.dockerfile
```diff
@@ -0,0 +1,11 @@
+ARG MONTAGU_GIT_ID="UNKNOWN"
```
hillalex, line 1: this is mostly just copied from previous location but github hasn't worked that out. the change is to the FROM bit - we now use a tagged image from docker hub because it gets built in a previous step and buildkite doesn't guarantee continuity of agents between steps (previously would just build a local image, call it "montagu-portal-shared-build-env" and inherit from that in the next step.)

.gitignore
```diff
@@ -9,4 +9,7 @@ react-prototype/npm-debug\.log
 app/config/orderlyweb/go_signal
```
EmmaLRussell, line 9: Can these app/ lines come out now?

hillalex:
> Nice, looks really good. It's a pity you have to dig around a bit more on BuildKite than on TC to see which tests are failing, though it makes the overall configuration steps more visible.

Yeah agree, I did actually find this https://www.npmjs.com/package/jest-buildkite-reporter so I'll try it out in a further PR

.gitignore
hillalex: The slight weirdness is that if you run `./scripts/run-development-dependencies.sh` from within the `app` dir, these things have to exist in the `app` directory because the `docker-compose` file has a reference to `$PWD`. But now I'm wondering if we can just set a different env var in the script that calls `docker compose`. Let me look into this.

hillalex: Ok this is much nicer :slightly_smiling_face:  https://github.com/vimc/montagu-webapps/pull/432/commits/37f3d187572bf425332a91b6c869dda554b87026

PR #436, opened 2020-08-21
vimc-4240 Exclude MONTAGU_TASK_QUEUE user from the user management page

EmmaLRussell:
Could have done in this in the api, but I figured it was valid to return this system user, we just don't want it being viewable in the portal.


hillalex (approved):
Seems like a good approach to me :+1:

PR #440, opened 2020-10-15
vimc-4325 don't show finished touchstones in active touchstone list

hillalex:
It's possible for a touchstone to have a previous version in un-finished status even though the most recent version is marked as "finished", e.g.
https://montagu.vaccineimpact.org/admin/touchstones/201907wue/

In this case the latest, finished, version should not be linked in the table of "active" touchstones: 
https://montagu.vaccineimpact.org/admin/touchstones/

nb: some of these cases are probably in error, we should also pick up this ticket to allow admins to easily mark touchstones as finished: https://mrc-ide.myjetbrains.com/youtrack/issue/VIMC-10

### DIFF
--- a/app/src/main/admin/components/Touchstones/List/TouchstoneList.tsx
+++ b/app/src/main/admin/components/Touchstones/List/TouchstoneList.tsx
@@ -16,7 +16,7 @@ export const TouchstoneListComponent: React.FunctionComponent<TouchstoneListProp
         <div className="mb-5 row">
             <div className="col-12">
                 <h4>Active touchstones</h4>
-                <div className="mb-2">The latest version of these touchstones is 'open' or 'in-preparation'.</div>
+                <div className="mb-2">These touchstones have at least one version that is 'open' or 'in-preparation'.</div>
                 <TouchstoneTable touchstones={props.active} showFinished={false}/>
             </div>
         </div>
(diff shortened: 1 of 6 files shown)


app/src/main/admin/components/Touchstones/List/TouchstoneList.tsx
```diff
@@ -16,15 +16,15 @@ export const TouchstoneListComponent: React.FunctionComponent<TouchstoneListProp
         <div className="mb-5 row">
             <div className="col-12">
                 <h4>Active touchstones</h4>
-                <div className="mb-2">The latest version of these touchstones is 'open' or 'in-preparation'.</div>
-                <TouchstoneTable touchstones={props.active}/>
+                <div className="mb-2">These touchstones have at least one version that is 'open' or 'in-preparation'.</div>
```
EmmaLRussell, line 19: ok, so in the case of touchstones which have a previous version unfinished, it still appears in the active list, but linking to the latest unfinished version?

EmmaLRussell (commented):
The code all looks fine to me. Can you tell what's going on with the buildkite failure?

hillalex:
> The code all looks fine to me. Can you tell what's going on with the buildkite failure?

Looks like the latest version of webpack-cli doesn't support node 8, which our build image is based on. Updating this base image is probably the way to go: https://github.com/vimc/node-docker

EmmaLRussell:
Right, and that change will have to go through the derived node-docker-openjdk image which the webapps image uses.

> Looks like the latest version of webpack-cli doesn't support node 8, which our build image is based on. Updating this base image is probably the way to go: https://github.com/vimc/node-docker

hillalex:
> Right, and that change will have to go through the derived node-docker-openjdk image which the webapps image uses.
> 
> > Looks like the latest version of webpack-cli doesn't support node 8, which our build image is based on. Updating this base image is probably the way to go: https://github.com/vimc/node-docker

I'm just doing this now, running changes through buildkite :+1:

hillalex:
Ok, update to node 10 has fixed the build. Once this is merged we can point everything back to master:
https://github.com/vimc/node-docker/pull/2

EmmaLRussell (approved):

PR #443, opened 2020-11-10
vimc-4378 add free text entry field and help text to upload coverage form

hillalex:
This adds the "description" field to the upload coverage form, along with some help text including an explanation of the variables in the template, as suggested by Katy.


app/src/main/admin/components/Touchstones/Coverage/CoverageVariablesPage.tsx
```diff
@@ -0,0 +1,53 @@
...
+                <li>country: the ISO code for this country</li>
+                <li>activity_type: "campaign" or "routine"</li>
+                <li>gavi_support: true or false - whether or not GAVI funded this program</li>
+                <li>year: one of {new Date().getFullYear()} - 2030</li>
+                <li>age_first: the lowest age that received this vaccine</li>
+                <li>age_last: the highest age that received this vaccine</li>
+                <li>gender: "female", "male", or "both"</li>
+                <li>target: the number of people in the target population</li>
+                <li>coverage: the fraction of the target population covered</li>
+                <li>subnational: true or false - was this distributed at a subnational level?</li>
```
hillalex, line 46: I went ahead and included "subnational" here even though it's not in the API yet, just so that when those tickets go in we don't have to do another update to the portal.

app/src/main/admin/components/Touchstones/Coverage/UploadCoverage.tsx
```diff
@@ -25,20 +26,49 @@ class UploadCoverageComponent extends React.Component<UploadCoverageProps, Uploa
...
+                            data sources
+                        </li>
+                        <li>
+                            things to note
+                        </li>
+                        <li>
+                            anything that could affect usage
+                        </li>
+                    </ul>
+                    Please see <InternalLink href={"coverage-variables"}>here</InternalLink> for information about how
```
EmmaLRussell, line 49: I think there's something up with this link / the breadcrumbs. If you click back and forth a bit, this link seems to get broken (points to touchstone/coverage/coverage-variables instead of touchstone/coverage-variables) and the same if you click on the breadcrumb link in the variables page itself.

app/src/main/admin/components/Touchstones/Coverage/CoverageVariablesPage.tsx
```diff
@@ -0,0 +1,53 @@
...
+                    RCV2,
+                    Rota,
+                    Rubella,
+                    Typhoid,
+                    YF
+                </li>
+                <li>country: the ISO code for this country</li>
+                <li>activity_type: "campaign" or "routine"</li>
+                <li>gavi_support: true or false - whether or not GAVI funded this program</li>
+                <li>year: one of {new Date().getFullYear()} - 2030</li>
```
EmmaLRussell, line 40: Isn't it this year + 1?

hillalex: Oh yeah you're right

app/src/main/admin/components/Touchstones/Coverage/UploadCoverage.tsx
hillalex: Ah, my bad, mismatch between breadcrumbs and router/link. It should all be `coverage/coverage-variables`, will update

hillalex:
The breadcrumbs should match the route/link now, although I couldn't actually replicate what you're describing with clicking back and forth so could you confirm that things are behaving?

EmmaLRussell:
Yeah sorry, that was a bit vague. These changes look good, but still getting Page not found sometimes. This seems to be repeatable:
- Go to coverage page
- Click on breadcrumb for the coverage page to reload it
- Click on the 'here' link

It seems to be linked to whether or not there's a / at the end of the current coverage url. The breadcrumb adds one, but you can get the same behaviour if you browse straight to coverage/. The url loaded by the 'here' link in that case seems to acquire an extra 'coverage' i.e. 'coverage/coverage/coverage-variables'

> The breadcrumbs should match the route/link now, although I couldn't actually replicate what you're describing with clicking back and forth so could you confirm that things are behaving?

hillalex:
> It seems to be linked to whether or not there's a / at the end of the current coverage url. The breadcrumb adds one, but you can get the same behaviour if you browse straight to coverage/. The url loaded by the 'here' link in that case seems to acquire an extra 'coverage' i.e. 'coverage/coverage/coverage-variables'
> 

Cheers, got it. Seems we need to use the absolute path in the `InternalLink` - updated to do so now.

PR #444, opened 2020-11-30
vimc-4391 show progress towards coverage uploads 

hillalex:
* Add new models to generated types
* Fetch uploaded coverage metadata on page load and also after every successful coverage upload
* Display metadata about uploaded coverage on page

(see mockups attached to YT ticket) 

Looks like a huge changeset, but includes test data csv which can be used with any of the touchstones in the test dataset when running locally. Browse to any touchstone then append `/coverage` to url to see the coverage upload page (link to the page not yet implemented).


app/README.md
```diff
@@ -16,6 +16,17 @@
 6. Run `./scripts/run-development-dependencies.sh` to run Montagu
    with a shared key, and with test data.
 
+## Generating Typescript classes from Kotlin classes
```
hillalex, line 19: Suprisingly I couldn't find this already documented

app/README.md
```diff
@@ -28,12 +39,16 @@ If you need more rules to check against, add them in file tslint.json, under sec
 3. `npm test -- -t "foo bar"` runs just the individual test called "foo bar".
 
 ## Integration tests
-Run `npm run integration-test` to run all integration tests. The version of
-the API that tests are run against is stored in `./config/api_version`.
+`npm run integration-test` runs all integration tests. The version of
+the API that tests are run against is stored in `./config/api_version`. 
+
+*NB be wary about running integration tests directly in your local dev environment. We have scripts which set up some
```
EmmaLRussell, line 45: Documentation update that would have saved me a bit of time!

mwoodbri (commented):
Sorry... took me a while to refresh my React knowledge! So, the code all looks good. Just a question over test coverage. And it appears to work well having tested manually but I noticed that it doesn't exactly follow mock-up from ticket - so just wanted to check whether that was by design. In particular:

- If I upload the same coverage file twice (or a subset of it) then I only see one message at the top of the page, for the most recent upload date ("A coverage set for YF was uploaded on Wed Dec 09 2020, 09:51 by test.user"). The mockup indicates that there should be two? This may just be because I'm uploading similar data.
- It says "A coverage set" rather than "Routine coverage..." or "Campaign coverage..." as indicated in the screenshots. I guess this is simply because the uploaded data can include both, so it doesn't actually make sense (or is even possible to) include that in the message.

EmmaLRussell:
Excellent point! I think you will probably only see one message for each coverage set uploaded (even if you upload it twice), but since a coverage set is specific to a vaccine and activity type, you should be able to see a distinction between those (even if they were both uploaded in the same file). I'll have a dig around and see why we're not getting the activity type passed back, might require a small change to the API. 

> Sorry... took me a while to refresh my React knowledge! So, the code all looks good. Just a question over test coverage. And it appears to work well having tested manually but I noticed that it doesn't exactly follow mock-up from ticket - so just wanted to check whether that was by design. In particular:
> 
>     * If I upload the same coverage file twice (or a subset of it) then I only see one message at the top of the page, for the most recent upload date ("A coverage set for YF was uploaded on Wed Dec 09 2020, 09:51 by test.user"). The mockup indicates that there should be two? This may just be because I'm uploading YF data twice - I don't think I can use an other vaccine for this touchstone version(?)
> 
>     * It says "A coverage set" rather than "Routine coverage..." or "Campaign coverage..." as indicated in the screenshots. I guess this is simply because the uploaded data can include both, so it doesn't actually make sense (or is even possible to) include that in the message.

app/src/main/admin/services/CoverageService.ts
```diff
@@ -1,9 +1,14 @@
...
-export class UploadCoverageService extends AbstractLocalService {
+export class CoverageService extends AbstractLocalService {
     uploadCoverage(touchstoneId: string, data: FormData) {
         return this.setOptions({notificationOnError: false})
             .post(`/touchstones/${touchstoneId}/coverage/`, data);
     }
+
+    fetchCoverageMetadata(touchstoneId: string) {
+        return this.setOptions({notificationOnError: false})
+            .get(`/touchstones/${touchstoneId}/coverage/meta/`);
```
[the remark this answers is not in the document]
EmmaLRussell: Yeah, there's an integration test but I forgot to add a unit test.

EmmaLRussell:
I've implemented adding activity type to the returned metadata in these branches:
https://github.com/vimc/montagu-webmodels/pull/95
https://github.com/vimc/montagu-api/pull/454

This branch is currently targeting vimc-4465 of the api for dev testing, in order to pick up these changes - this can be reverted to 'master' when those are merged. 

Still very sightly different to the mockup as configured display name for routine is 'Routine vaccination', not just 'Routine' (and similarly for campaign) but this provides the missing activity type information.

mwoodbri (approved):
Looks great.

Tested by uploading `manualTestData/upload_coverage.csv` twice: once in full and then with just campaign activity. Alerts correctly showed (different) upload dates.

PR #445, opened 2020-12-03
vimc-4433 Restrict coverage route to users with coverage-write permission

EmmaLRussell:
Interestingly, what this ticket has revealed is that, in fact, we do not currently lock down any other routes in the portals! We just strategically hide links to the pages, and the user who stumbles on a route they shouldn't find will just have a bad time, because the API which is serving the data will return 401s when the portal requests it. 

I think we got away with this before, when the modellers accessing the contrib portal, and admin users accessing the admin portal would probably all have much the same sets of permissions, but now that we'll be opening the Admin portal to GAVI users with restricted privileges to upload coverage only we should probably be more stringent. 

Making this ticket a proof of concept for putting this in for the coverage route, and have made another ticket for covering all the routes (https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2034). The router will not add routes the user is not permitted to access, and the user will see the 'Page not found' screen if they attempt to browse to the route. 

I've rearranged the top-level components a little, so that the router can now access properties in the state directly rather than receiving them as props - this means less boiler plate when we add switching route addition on the user's permissions for the other pages. I've also added canUploadCoverage to the auth state - we will use this from the touchstone page when we add the coverage link. Again, this can be a pattern we repeat when we conditionally add the other routes - currently many components calculate their own canDoSomething properties when deciding whether to add links, but we can move those into the state to be accessible from the router too. 

This can be manually tested locally by removing the line `$here/cli.sh addRole test.user coverage-provider` from `add-test-accounts-for-integration-tests.sh` and rerunning dev dependencies. Browse to a touchstone version page, and append `/coverage` to attempt to access the coverage page.

### DIFF
--- a/app/src/main/shared/reducers/authReducer.ts
+++ b/app/src/main/shared/reducers/authReducer.ts
@@ -10,6 +10,7 @@ export interface AuthState {
     isAccountActive: boolean;
     isModeller: boolean;
     errorMessage?: string;
+    canUploadCoverage: boolean;
 }
 
 export const initialAuthState: AuthState = {
@@ -18,7 +19,8 @@ export const initialAuthState: AuthState = {
     bearerToken: null,
     permissions: [],
     isAccountActive: false,
-    isModeller: false
+    isModeller: false,
+    canUploadCoverage: false
 };
 
 export interface AuthStateOptions {
@@ -29,15 +31,18 @@ export interface AuthStateOptions {
     modellingGroups: string[]
 }
 
+const permissionsInclude = (permissions: string[], perm: string) => permissions.some((x: string) => x == perm);
+
 export function loadAuthState(options: AuthStateOptions): AuthState {
     return {
         loggedIn: options.loggedIn,
         bearerToken: options.bearerToken,
-        isAccountActive: options.permissions.some((x: string) => x == "*/can-login"),
+        isAccountActive: permissionsInclude(options.permissions,"*/can-login"),
         isModeller: options.modellingGroups.length > 0,
         username: options.username,
         permissions: options.permissions,
-        modellingGroups: options.modellingGroups
+        modellingGroups: options.modellingGroups,
+        canUploadCoverage: permissionsInclude(options.permissions, "*/coverage.write")
     }
 }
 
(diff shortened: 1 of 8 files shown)


app/src/test/admin/components/AdminAppTests.tsx
```diff
@@ -5,11 +5,11 @@ import {AdminApp, AdminAppComponent} from "../../../main/admin/components/AdminA
 import {AdminRouter} from "../../../main/admin/components/AdminRouter";
 
 describe("AdminApp", () => {
-    it("passes through loggedIn to router", () => {
-        const rendered = shallow(<AdminAppComponent loggedIn={ true } />);
+    it("passes through history to router", () => {
+        const rendered = shallow(<AdminAppComponent loggedIn={ true } history={ "TEST HISTORY" as any } />);
```
mwoodbri, line 9: Maybe remove `loggedIn` here? I know it's now `undefined` because it's derived from state but it's slightly confusing to be testing a legacy prop.

app/src/main/shared/reducers/authReducer.ts
```diff
@@ -18,7 +19,8 @@ export const initialAuthState: AuthState = {
     bearerToken: null,
     permissions: [],
     isAccountActive: false,
-    isModeller: false
+    isModeller: false,
+    canUploadCoverage: false
```
mwoodbri, line 23: Feel like a bit of a code-smell that we're providing both `permissions` and `modellingGroups` *and* derived fields such as `isAccountActive`, `isModeller` and `canUploadCoverage`. I worry that e.g. changing the definition of a `isModeller` could be difficult if code elsewhere is inferring it from `modellingGroups`. But I don't think it's specific to this PR and I'm not fully aware of broader context!

mwoodbri (approved):
LGTM. One comment and one potential change but otherwise happy as-is.

Also tested, as follows:

1. Visited http://localhost:5000/admin/touchstones/op-2017/op-2017-2/coverage/
1. Saw coverage upload page
1. Edited `app/scripts/add-test-accounts-for-integration-tests.sh` as described
1. Revisited http://localhost:5000/admin/touchstones/op-2017/op-2017-2/coverage/
1. Saw “Page not found”

app/src/main/shared/reducers/authReducer.ts
EmmaLRussell: Agreed. I want to centralise the isTypeOfUser and canDoSomething properties into the auth state as part of the work to lock down the other pages. Then the components which are currently working that stuff out themselves can use those properties in the auth state, and the router can use them too. We should be able to take the permissions list out of the state as a property at that point. I may try tackling this for our Weds afternoon maintenance session this week...

PR #446, opened 2020-12-08
vimc-4435 Show link to coverage upload page if user has permission

EmmaLRussell:
Added the coverage page link the touchstone details component (where user has corrrect permission). Also refactored tests to test the wiring up of the states to the props as well as correct rendering.

This is ready to review but should not be merged until https://github.com/vimc/montagu-webapps/pull/444 is merged.


mwoodbri (approved):
Looks good and works well - link disappears from page (and direct access to page denied) when relevant permission removed from `montagu.user_group_role`. Tests also much tidier now!

PR #449, opened 2020-12-10
vimc-4461 Put canDoThing properties on auth state rather than components

EmmaLRussell:
Remove permissions as a property on auth state, and put all knowledge of individual permissions and the abilities they confer into the state, and use these properties in the components. 

This is preparatory for using these same properties in the router to hide pages from users who should not have access to them

### DIFF
--- a/app/README.md
+++ b/app/README.md
@@ -96,7 +96,7 @@ management.
 
 # User permissions
 
-Users can access some areas of the portals dependent on the permissions they possess. Raw user permissions from the [Montagu
+Users can access some areas of the admin portal dependent on the permissions they possess. Raw user permissions from the [Montagu
 database](https://github.com/vimc/montagu-db) are converted into properties on the `AuthState` interface, indicating what
 the user is allowed to do in the portals. These properties are used by the components to determine whether links and
 controls should be shown to the user. In addition, the [Montagu API](https://github.com/vimc/montagu-api) will prevent 
--- a/app/README.md
+++ b/app/README.md
@@ -93,3 +93,27 @@ Short name: `admin`
 This portal allows administrative staff to set up touchstones with all the
 necessary input data, responsibilities and recipes. It's also where we do user
 management.
+
+# User permissions
+
+Users can access some areas of the portals dependent on the permissions they possess. Raw user permissions from the [Montagu
+database](https://github.com/vimc/montagu-db) are converted into properties on the `AuthState` interface, indicating what
+the user is allowed to do in the portals. These properties are used by the components to determine whether links and
+controls should be shown to the user. In addition, the [Montagu API](https://github.com/vimc/montagu-api) will prevent 
+the user taking any actions for which they do not have permissions.
+
+The `AuthState` properties, and their corresponding permissions are:
+
+| Property | Permission | Used in
+| --- | --- | --- |
+| `canDownloadCoverage` | `*/coverage.read` | Touchstone scenarios page
+| `canUploadCoverage` | `*/coverage.write` | Touchstone coverage upload page
+| `canViewGroups` | `*/modelling-groups.read` | Main menu 
+| `canViewTouchstones` | `*/touchstones.read` | Main menu
+| `canViewUsers` | `*/users.read` | Main Menu
+| `canReadRoles` | `*/roles.read` | User details page
+| `canWriteRoles` | `*/roles.write` | User details page
+| `canCreateUsers` | `*/users.create` | Users page (CreateUsersSection component)
+| `canCreateModellingGroups` | `*/modelling-groups.write` | Modelling Groups page (CreateModellingGroupSection component)
+| `canManageGroupMembers` | `*/modelling-groups.manage-members` | Modelling Group details page (several components)
+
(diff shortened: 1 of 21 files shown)


app/src/main/shared/reducers/authReducer.ts
```diff
@@ -40,9 +56,18 @@ export function loadAuthState(options: AuthStateOptions): AuthState {
...
+
+        canDownloadCoverage: permissionsInclude(options.permissions,"*/coverage.read"),
+        canUploadCoverage: permissionsInclude(options.permissions, "*/coverage.write"),
+        canViewGroups: permissionsInclude(options.permissions,"*/modelling-groups.read"),
+        canViewTouchstones: permissionsInclude(options.permissions,"*/touchstones.read"),
+        canViewUsers: permissionsInclude(options.permissions, "*/users.read"),
+        canReadRoles: permissionsInclude(options.permissions,"*/roles.read"),
+        canWriteRoles: permissionsInclude(options.permissions,"*/roles.write"),
+        canCreateUsers: permissionsInclude(options.permissions,"*/users.create"),
+        canCreateModellingGroups: permissionsInclude(options.permissions,"*/users.create"),
```
EmmaLRussell, line 69: We're actually using the same permission for this as for canCreateUser, but the components don't need to know this.

app/src/test/shared/reducers/AuthReducersTests.ts
```diff
@@ -108,11 +128,50 @@ describe ('loadAuthState tests', () => {
...
         });
-        expect(result.canUploadCoverage).toBe(true);
+    };
+
+    it('sets canUploadCoverage if coverage write permission present', () => {
+        expect(loadAuthStateWithPermission("*/coverage.write").canUploadCoverage).toBe(true);
+    });
+
+    it('sets canDownloadCoverage if coverage read permission present', () => {
+        expect(loadAuthStateWithPermission( "*/coverage.write").canUploadCoverage).toBe(true);
```
mwoodbri, line 143: Copy-paste dupe?

app/src/main/shared/reducers/authReducer.ts
mwoodbri: Can you provide a bit more context for the apparent anomaly of `modelling-groups.read` vs `users.create`?

mwoodbri (changes requested):
All looks good besides one typo to resolve (and a question for my benefit!).

Note that I haven't tested this manually but happy to do so if that would be valuable/helpful.

EmmaLRussell:
> Can you provide a bit more context for the apparent anomaly of modelling-groups.read vs users.create?

You mean modelling-groups.create? It's a bit smelly isn't it? I'm not really sure, I think it was before my time. I suspect there was a frustration at the proliferation of permissions, and someone decided that having the authority to create users was equivalent to that to create modelling groups, and to reuse that permission for modelling group creation. Possibly justified by the reasoning that modelling groups are like user groups. 

This could be made less smelly by renaming users.create to something more generic, which would be a fairly delicate update to do on the existing live data. I expect it was originally decided that change wasn't worth the risk. I think corralling that ugliness into this prop is a slight improvement. Documenting the perms and how they come into play in the README would also help?

mwoodbri:
> > Can you provide a bit more context for the apparent anomaly of modelling-groups.read vs users.create?
> 
> You mean modelling-groups.create? It's a bit smelly isn't it? I'm not really sure, I think it was before my time. I suspect there was a frustration at the proliferation of permissions, and someone decided that having the authority to create users was equivalent to that to create modelling groups, and to reuse that permission for modelling group creation. Possibly justified by the reasoning that modelling groups are like user groups.
> 
> This could be made less smelly by renaming users.create to something more generic, which would be a fairly delicate update to do on the existing live data. I expect it was originally decided that change wasn't worth the risk. I think corralling that ugliness into this prop is a slight improvement. Documenting the perms and how they come into play in the README would also help?

Ah, makes sense. Yes, I was referring to lack of `modelling-groups.create`. If it doesn't exist them some documentation about permissions would help, even if it just a) links to the canonical definitions within the code/database and a) explains the re-use of `users.create`. I agree that attempting to rename it is folly.

EmmaLRussell:
Ha, turned out my theory was wrong about the origin of the `canCreateModellingGroup`  and `*/users.create` issue - when I had another look at how it was used, I got suspicious that the historic re-use of the users create permission was another copy paste error, as it's a close match for the Users Create component. 

So I checked  the [api](https://github.com/vimc/montagu-api/blob/master/src/app/src/main/kotlin/org/vaccineimpact/api/app/app_start/route_config/ModellingGroupRouteConfig.kt), and, sure enough, there is actually a `*/modelling-group.write` permission which controls whether you can create a modelling group. So this is a historic bug which we've got away with because our admin users have both permissions...

I'll fix that up in this branch, and keep on with adding a bit of documentation anyway.

app/README.md
```diff
@@ -93,3 +93,27 @@ Short name: `admin`
...
+| `canUploadCoverage` | `*/coverage.write` | Touchstone coverage upload page
+| `canViewGroups` | `*/modelling-groups.read` | Main menu 
+| `canViewTouchstones` | `*/touchstones.read` | Main menu
+| `canViewUsers` | `*/users.read` | Main Menu
+| `canReadRoles` | `*/roles.read` | User details page
+| `canWriteRoles` | `*/roles.write` | User details page
+| `canCreateUsers` | `*/users.create` | Users page (CreateUsersSection component)
+| `canCreateModellingGroups` | `*/modelling-groups.write` | Modelling Groups page (CreateModellingGroupSection component)
+| `canManageGroupMembers` | `*/modelling-groups.manage-members` | Modelling Group details page (several components)
+
```
mwoodbri, line 119: This is great - thanks

mwoodbri (approved):

PR #450, opened 2021-01-28
vimc-4545 Users without permission should get 404s on user list and details pages

EmmaLRussell:
Does not include these pages in the router when user does not have */users.read permission. 

Can be tested manually by commenting out `$here/cli.sh addRole test.user admin` in the script `add-test-accounts-for-integration-tests.sh` before running the dependencies - browsing to these pages should now show a 'Page not found' page rather than 'You do not have sufficient permissions' errors and endless spinner.

### DIFF
--- a/app/src/test/admin/components/AdminRouterTests.tsx
+++ b/app/src/test/admin/components/AdminRouterTests.tsx
@@ -14,6 +14,8 @@ import {createMockAdminStore} from "../../mocks/mockStore";
 import {RecursivePartial} from "../../mocks/mockStates";
 import {AdminAppState} from "../../../main/admin/reducers/adminAppReducers";
 import {CoveragePage} from "../../../main/admin/components/Touchstones/Coverage/CoveragePage";
+import {UsersListPage} from "../../../main/admin/components/Users/List/UsersListPage";
+import {UserDetailsPage} from "../../../main/admin/components/Users/SingleUser/UserDetailsPage";
 
 describe("AdminRouter", () => {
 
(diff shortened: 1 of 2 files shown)


app/src/test/admin/components/AdminRouterTests.tsx
```diff
@@ -49,4 +51,22 @@ describe("AdminRouter", () => {
...
+        expect(rendered.find(UsersListPage)).toHaveLength(0);
+    });
+
+    it("includes user page only when current user can view users", () => {
+        const userRoute = "/users/test-user/";
+        let rendered = renderComponent({auth: {loggedIn: true, canViewUsers: true}}, userRoute);
+        expect(rendered.find(UserDetailsPage)).toHaveLength(1);
+
+        rendered = renderComponent({auth: {loggedIn: true, canViewUsers: false}}, userRoute);
+        expect(rendered.find(UsersListPage)).toHaveLength(0);
```
mwoodbri, line 70:
```suggestion
        expect(rendered.find(UserDetailsPage)).toHaveLength(0);
```
Copypasta?

mwoodbri (changes requested):
Does what it says on the tin! One fix to make I think.

A couple of observations, which I guess are just a result of running the admin instance of the app?
- Even without the `admin` role, users can see "admin" routes e.g. `/admin/touchstones/`
- Visiting `/` and `/admin/` both display the "Main menu"

app/src/test/admin/components/AdminRouterTests.tsx
EmmaLRussell: oops!

EmmaLRussell:
>     * Even without the `admin` role, users can see "admin" routes e.g. `/admin/touchstones/`
> 

I think that's because the view touchstones permission is also in one of the other roles (modeller probably)

>     * Visiting `/` and `/admin/` both display the "Main menu"

That's a result of the way the app is run locally where one of the portals is root I believe, should be fine.

mwoodbri (approved):

PR #451, opened 2021-01-28
vimc-4546 Add mq and flower to docker compose file

EmmaLRussell:
Adds redis message queue and flower to docker compose for dev dependencies. Without them, an error is shown on burden estimate upload when the API tries to put a message on the queue to kick off a diagnostic report. 

This error probably shouldn't be shown to user anyway as it indicates a problem with montagu infrastucture rather than a problem with the user's upload. This would be a fix in API. But this helps with testing for now.

Test this fix by uploading the burden estimates csv file in the attached zip in the contribution portal, for op-2017-2 touchstone, routine scenario. No error should be shown.

[dummyDEVcentral-burden-template.op-2017-2.description (1).zip](https://github.com/vimc/montagu-webapps/files/5888385/dummyDEVcentral-burden-template.op-2017-2.description.1.zip)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -28,14 +28,10 @@ services:
       - $ROOT/demo:/orderly
       - $ROOT/config/orderlyweb:/etc/orderly/web
   mq:
-    image: redis
-    ports:
-      - "6379:6379"
+    image: redis:6
   flower:
     image: mher/flower:0.9.5
-    ports:
-      - "5555:5555"
     environment:
       - FLOWER_PORT=5555
-      - CELERY_BROKER_URL=redis://guest@montagu_mq_1//
-      - CELERY_RESULT_BACKEND=redis://guest@montagu_mq_1/0
+      - CELERY_BROKER_URL=redis://guest@mq//
+      - CELERY_RESULT_BACKEND=redis://guest@mq/0
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -27,3 +27,15 @@ services:
     volumes:
       - $ROOT/demo:/orderly
       - $ROOT/config/orderlyweb:/etc/orderly/web
+  mq:
+    image: redis
+    ports:
+      - "6379:6379"
+  flower:
+    image: mher/flower:0.9.5
+    ports:
+      - "5555:5555"
+    environment:
+      - FLOWER_PORT=5555
+      - CELERY_BROKER_URL=redis://guest@montagu_mq_1//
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -27,3 +27,15 @@ services:
     volumes:
       - $ROOT/demo:/orderly
       - $ROOT/config/orderlyweb:/etc/orderly/web
+  mq:
+    image: redis
+    ports:
+      - "6379:6379"
```
mwoodbri, line 33: Do these ports need to be exposed to the host machine? Shouldn't be necessary if only used for communication between containers defined in this file.

If do need to be exposed then consider "127.0.0.1:6379:6379" otherwise they're exposed to other hosts on the network.

docker-compose.yml
```diff
@@ -27,3 +27,15 @@ services:
...
+  flower:
+    image: mher/flower:0.9.5
+    ports:
+      - "5555:5555"
```
mwoodbri, line 37: As above, though may be worth exposing to host via `127.0.0.1:5555:5555` as Flower appears to have a web interface

docker-compose.yml
```diff
@@ -27,3 +27,15 @@ services:
     volumes:
       - $ROOT/demo:/orderly
       - $ROOT/config/orderlyweb:/etc/orderly/web
+  mq:
+    image: redis
```
mwoodbri, line 31: Do we need a (major) version tag?

docker-compose.yml
```diff
@@ -27,3 +27,15 @@ services:
...
+    ports:
+      - "6379:6379"
+  flower:
+    image: mher/flower:0.9.5
+    ports:
+      - "5555:5555"
+    environment:
+      - FLOWER_PORT=5555
+      - CELERY_BROKER_URL=redis://guest@montagu_mq_1//
```
mwoodbri, line 40: Should be able to use `redis://guest@mq//` to avoid hard-coding Compose project name and instance no.

Same for next line.

mwoodbri (approved):
Some Compose minutiae but nothing critical given that it's a dev setup.

Disregarding that it works very nicely. First time I've seen this bit of the app. I uploaded as single run estimates and got graph etc.

docker-compose.yml
EmmaLRussell: Yeah, good point. I think we tend to expose more ports in general than we need to, almost just as a reminder of which are the relevant ports on each container... It can be useful in deployed networks for diagnostics I guess, unlikely to be needed in these sorts of dev scripts. Good point on keeping them local, we should probably do that more too.

PR #464, opened 2021-07-13
VIMC-4775 Remove admin role assumption from responsibilities view

mwoodbri:
- Ensure that touchstone responsibility page renders correctly
  - If you have `responsibilities.review` permission (i.e. the typical situation of having the `admin` role) then you see the comments column and can add/edit comments
  - If you just don't have that permission (i.e. you have the `user` role) then the page loads regardless but you can't see the column. The endpoint itself is already protected.
- You can test this by running the app, visiting (e.g.) [this page](http://localhost:5000/admin/touchstones/op-2017/op-2017-2/responsibilities/) and checking you can add/edit comments. Then remove `admin` role from `test.user` and revisit the page:
  ```sh
  docker exec -it montagu_db_1 psql -Uvimc montagu -c "delete from user_group_role where role = 17"
  ```
- This PR also updates URL of commenting endpoint to match latest `montagu-api`
- Some further streamlining to how annotated responsibility sets are passed between state/components will follow in VIMC-4767

### DIFF
--- a/app/README.md
+++ b/app/README.md
@@ -116,4 +116,4 @@ The `AuthState` properties, and their corresponding permissions are:
 | `canCreateUsers` | `*/users.create` | Users page (CreateUsersSection component)
 | `canCreateModellingGroups` | `*/modelling-groups.write` | Modelling Groups page (CreateModellingGroupSection component)
 | `canManageGroupMembers` | `*/modelling-groups.manage-members` | Modelling Group details page (several components)
-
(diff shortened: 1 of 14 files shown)


EmmaLRussell (commented):
Looks great to me, works perfectly. Could add a test to AuthReducersTests that `canReviewResponsibilities` is correctly initialised from the permissions.

mwoodbri:
> Looks great to me, works perfectly. Could add a test to AuthReducersTests that `canReviewResponsibilities` is correctly initialised from the permissions.

Drat.. there's always one! Thanks - done.

EmmaLRussell (approved):

PR #466, opened 2021-08-04
VIMC-4809 Add guidance pages for touchstone 202108test

mwoodbri:
To test:
- Modify `default.ts` so that `is2021Touchstone` returns `true` for a touchstone existing on your system e.g. `op-2018-1` (or maybe create a `202108test` touchstone, though I couldn't do that for some reason)
- Build/run contrib app
- Visit http://localhost:5000/contribution/IC-Garske/responsibilities/op-2018-1/
- Contents of "Useful links" should look like documents attached VIMC-4809. Note that help links won't work locally without the relevant resources but do work on UAT/science/prod.

Notes:
- This guidance is displayed only for the `202108test` touchstone, rather than any touchstone beginning with `202108` (or just `2021`) because the text refers explicitly to practice runs - we expect further guidance to be issued for the real ones.
- I've assumed links to Montagu reports provided in documents will be accessible to relevant users. If not we could convert reports to PDF and attach.


app/src/main/contrib/components/Responsibilities/Guidance/content/ResponsibilityGuidanceModelInputsContent2021.tsx
```diff
@@ -0,0 +1,86 @@
...
+                included one no-vaccination and one default scenario. Depending on the disease, the default scenario may
+                contain both routine and campaign coverage sets, or just one of these.</p>
+            <p>There are no coverage sets included in the no vaccination scenario, and therefore this coverage file does
+                not contain any data apart from the header row.</p>
+            <p><b>gavi_support</b> can be ignored. All column values are ‘total’; this means either Gavi-funded or not
+                Gavi-funded.</p>
+            <p><b>age_range_verbatim</b> is a description of the age range. We have used this to
+                infer <b>age_first</b> and <b>age_last</b>. Where age_range_verbatim is ‘NA’, this indicates the default
+                age_first and age_last value for that vaccine.</p>
+            <p><b>coverage</b> shows the level of vaccination coverage</p>
```
EmmaLRussell, line 34: In the doc, this sentence continues "ranging from 0 (0%) to 1 (100%)."

app/src/main/contrib/components/Responsibilities/Guidance/content/ResponsibilityGuidanceModelOutputsContent2021.tsx
```diff
@@ -0,0 +1,103 @@
...
+            <div className="largeSectionTitle">Overview</div>
+            <p>The 2021 practice runs are not a requirement, but we encourage modellers to provide them to ensure their
+                model will work on Montagu for the full 2021 runs. For the practice runs, we expect only central
+                estimates, not stochastic estimates.</p>
+            <p>(For the full model runs, we will require both central and stochastic estimates.)</p>
+
+            <div className="largeSectionTitle">Countries</div>
+            <ul>
+                <li>JE groups: China, Pakistan, India</li>
+                <li>Men A & YF groups: Ethiopia, Nigeria</li>
```
EmmaLRussell, line 28: Pedantic point, and inherited from the original doc, but I think this would be more consistent with following line
```suggestion
                <li>Men A and YF groups: Ethiopia, Nigeria</li>
```

EmmaLRussell (commented):
Looks good to me, couple of tiny comments

EmmaLRussell (approved):

PR #472, opened 2021-09-22
vimc-4868 Add guidance for 202110gavi touchstone

EmmaLRussell:
Adds guidance for the 2021gavi touchstone, as pdfs, in the same manner as for the 201910gavi touchstone. Adds `is2021GaviTouchstone` method to settings to check for this, and renames the old `is2021Touchstone` method to `is2021TestTouchstone` to distinguish it.

For manual testing, the test settings point this method to the op-2018 touchstone, so you should see the 2021 gavi pdf links when you browse to that touchstone's responsibillities in local dev testing. 

The responsibilities page for any version of the touchstone should include links to these pdfs, which should be visible when click link. Also, should see relevant links, when browse to location where one would see guidance html pages in those touchstones which provide html guidance: `help/model-inputs/{touchstone id}` and  `help/model-outputs/{touchstone id}`


mwoodbri (approved):
Looks good... links from all three pages point at expected content, and content for other touchstones remains unchanged. Touchstone name as expected.

PR #473, opened 2021-09-22
vimc-4870 Remove obsolete text from responsibilities page

EmmaLRussell:
For stochastic touchstones, remove text implying that stochastic data will be uploadable via the portal 'in the future'.

Remove reference to montagu slack channel from default responsibility description contact section.


mwoodbri (approved):
Removed op-2018 from nonStochasticTouchstones locally and did a visual diff with 202110gavi-4 on science - relevant content removed as expected

